PR #202, opened 2015-05-11
Added averaging method "add"

dodogerstlin:
This is useful for summing up short exposure images from counting detectors. Only tested it with edf files created by our machinee, maybe also useful for others?


kif:
Would you mind calling it "sum" instead of "add" as it reminds directly the sum method of numpy.

Thanks.

dodogerstlin:
not at all! i changed it by now i think..

PR #276, opened 2016-05-27
Cleanup

vallsv:
Here is few smalls clean up
Tested with testsuite (without GPU suite)


kif:
looks good. tests passed on CI (Travis & AppVeyor)
processor.py will be removed, replaced with worker.py

Issue #283, opened 2016-06-08
remaing files due to the unit test


picca:
Hello, while preparing the pyFAI 0.12 debian package I found these files in the source directory.

all_testimages.json
fai_cfg.json
lima_cfg.json
progile.log

it would be nice if you could  remove these file during the teardown

vallsv:
Thanks for the feed back. I will try to fix few of them.

vallsv:
I fixed all_testimages.json and profile.log in patches above (there are only 2 paches)

fai_cfg.json and lima_cfg.json come from io.py (lines 253 and 254).
It looks like this files was generated for debug purpose only. What do you think about @kif? We only have to remove this lines?

vallsv:
Everything is fixed. It will be merged soon.
Thanks again for the feed back.

PR #287, opened 2016-06-08
Avoid to generate profile.log default tests (improve #283)

vallsv:
- Avoid generateing the log files for default tests
- Ignore the file in other cases


kif:
looks OK

PR #309, opened 2016-06-15
Fix small issues

vallsv:
- Few changes/cleanup on the bootstrap (4 first commits)
- Update of .gitignore (1 commit)

Other commits fix little things to allow to generate man files without manual edition
- Use ArgumentParser for all scripts
- Update make_man
- Add missing man/pdf for benchmark
- Update other man/pdf files
- man/pdf files are still generated into ./man


kif:
Looks good to me.

PR #314, opened 2016-06-21
Integrate silx mask

### LINKED ISSUE #311: Use Silx mask instead of PyMca one
It is only used in drawMask_pymca

Then... should we rename it?
- ```pyFAI-drawmask```?
- And create a kind of alias for ```drawMask_pymca```?

vallsv:
- Rename  drawMask_pymca to pyFAI-drawmask
- Use silx instead of PyMCA if possible (close #311)
- Rework the main window


scripts/pyFAI-drawmask
```diff
@@ -0,0 +1,220 @@
...
+
+"""
+pyFAI-drawmask
+
+Use silx or PyMca module to define a mask
+"""
+
+__authors__ = ["Jerome Kieffer", "Valentin Valls"]
+__contact__ = "Jerome.Kieffer@ESRF.eu"
+__license__ = "GPLv3+"
```
kif: One comment: when a file has been modified only by the silx-team (full ESRF copyright) it would be great to switch it to MIT license and declare the file in copyright accordingly.

PR #329, opened 2016-06-30
Project quality

vallsv:
- Most of it is to move directories and rename few files.
- One contribution to improve unittest
- and one to clean up dup logging configuration


kif:
looks OK to me

PR #335, opened 2016-06-30
fix #333

### LINKED ISSUE #333: create hexversion from version_info
to be usable from outside. It should also work with simple tuples

kif:
this is the preferred way to calculate the hexversion


kif:
do not merge yet ...

kif:
now it should be ready to merge.

kif:
unles some test don't pass

kif:
ready for review

PR #350, opened 2016-07-11
Fix issue 64

vallsv:
Here is a big request but checking it commit by commit should be clear.

The idea was to fully avoid to use Qt by using a simple writer

- The 6 first commits prepare the code to be able to use a writer from the worker
- The last 4 rework the worker and the interate scripts

- 6th remove dup code, but if you dont like the idea, we can drop that one
- On the 9th, i create a new package "util" for everything without dependency with pyfai. I guess we can split utils.py there. If you dont like the idea i can also rework that.


vallsv:
Have to way the CI

pyFAI/azimuthalIntegrator.py
```diff
@@ -3266,38 +3493,24 @@ def save1D(self, filename, dim1, I, error=None, dim1_unit=units.TTH,
         @type error: numpy.ndarray or None
         @param dim1_unit: the unit of the dim1 array
         @type dim1_unit: pyFAI.units.Unit
-        @param dark: save the darks filenames (default: no)
-        @type dark: ???
-        @param flat: save the flat filenames (default: no)
-        @type flat: ???
+        @param has_dark: save the darks filenames (default: no)
```
kif: Were dark correction applied ?

kif:
Hi Valentin,

Thanks for the pull request.

Just a few comments:
wouln't it be meaningful to define the 3 classes IntegrateResult in another module ? the azimuthal integrator module is already too large and they have no "synergy" being there.
Moreover it would be nice to have also an object returned when "all=True" with the same attributes as IntegrateResult (but not being a tuple, as it is a dict

Overall the restructuring looks good. As some methods changed signatures, the docstrings are no more correct.

vallsv:
Sure,

- let me know in which module we can put results? resulttuples.py? results.py? I think the current location is also nice cause only interate1d and 2d should generate this objects (I use protected setter for example). But moving it makes sens too.
- for all=True i would prefer to force people to not use it. Then i prefer to see it like a compatibility code (just the minimum), and people have to use the new way if they need more. But it is up to you. I you prefer,  Ill fix the dictionary.
- I see the pbm of the docstring, ill fix it too

kif:
On Tue, 12 Jul 2016 04:49:59 -0700

> Sure,
> 
> - let me know in which module we can put results? resulttuples.py? results.py? I think the current location is also nice cause only interate1d and 2d should generate this objects (I use protected setter for example). But moving it makes sens too.
> - for all=True i would prefer to force people to not use it. Then i prefer to see it like a compatibility code (just the minimum), and people have to use the new way if they need more. But it is up to you. I you prefer,  Ill fix the dictionary.
> - I see the pbm of the docstring, ill fix it too

It would probably make sense to put them in the same module as the "Enum" class used (and defined in units)
We could have a module with "containers" if there is no collision with anything in the python distribution it could be OK ?

Any comment about the name ?

Cheers,
-- 
Jérôme Kieffer
tel +33 476 882 445

vallsv:
If CI passes, i think it is ok.
If you really want to extand the dict when all=True, we can open a bug. I can fix it into another patch.
.

PR #361, opened 2016-07-21
Fix appveyor compilation for python 3.4 and 3.5

vallsv:


ci/appveyor.yml
```diff
@@ -32,48 +32,67 @@ environment:
...
-        # Python 3.4: disabled for now, investigate why it crashes
-        #- PYTHON_DIR: "C:\\Python34-x64"
-        #  PYTHON_VERSION: "3.4.3"
-        #  PYTHON_ARCH: "64"
-        #  DISTUTILS_USE_SDK: "1"
+        # Python 3.4
+        - PYTHON_DIR: "C:\\Python34-x64"
+          PYTHON_VERSION: "3.4.3"
+          PYTHON_ARCH: "64"
+          DISTUTILS_USE_SDK: "1"
```
t20100: DISTUTILS_USE_SDK seems unused

t20100:
DISTUTILS_USE_SDK variable seems useless.
Otherwise, it looks good to me.

vallsv:
We fixed that with Jerome from the appveyor doc https://packaging.python.org/appveyor/
Then i thought it was accessed by the python library from the env.

But thanks for the warn. I think in our case it is really useless.
Let me rework a little the patch using this appveyor doc
cause the MSVC SetEnv.cmd must be called in difference cases: Python 3.3 x64 or Python 3.4 x64 only

vallsv:
Ok, this one should be the right one. If appveyor is agree

PR #364, opened 2016-07-25
Ignore last version of setuptool for appveyor

vallsv:
- All version 24.* are not working
- The last version 25.0 is also not working


vallsv:
Merged in https://github.com/silx-kit/pyFAI/pull/366

Issue #368, opened 2016-07-27
pyFAI-calib is not compatible with Python 3


sghose:
It seems that there is no `input` function specified in the .`utils` module if running python 3

vallsv:
Hi. ```input``` is defined like that at line 95 of ```utils.py``` on the master

```
input = raw_input
```

It allow to have compatible code between Python 2 and Python 3.
Are you talking about a specific version?

vallsv:
I think you are right. The idea is to use the builtin function

It is loaded like that in one of the test files
```
try:
    from ..utils import input
except ImportError:
    pass
```

Then in this case you can use ```input``` in the module, but you are right, there is no ```input``` in the utils module in python 3. I think we should simplify ```input/raw_input``` calls, using ```six``` or our own input. Right now there is no abstraction. What do you think @kif?

But do you have an issue somewhere? The use of ```utils.input``` should only be internal to pyfai.

CJ-Wright:
So the issue we were facing was with a pip install of pyFAI into a python 3 environment, the input issue created an error when using pyFAI-calib

CJ-Wright:
I think that @chiahaoliu's fork has some of these issues addressed

vallsv:
I can reproduce the error with the master.
```
  File "/home/valls/workspace/pyfai.git/build/scripts-3.4/pyFAI-calib", line 48, in <module>
    from pyFAI.calibration import Calibration
  File "/home/valls/workspace/pyfai.git/build/lib.linux-x86_64-3.4/pyFAI/calibration.py", line 65, in <module>
    from .utils import measure_offset, expand_args, \
ImportError: cannot import name 'input'
```

I will try to reproduce it with the previous release.
If you can use the master from git, this issue will be fixed today. If you need to wait a release it will take a little longer.

vallsv:
There is other problems relative to Python 3. Then i rename this bug.

vallsv:
Git master is now patched. Let me know if you have anymore problems with it.

I think it is a good idea to backport that to the branch 0.12. But yet i dont know if we will do that.

Thanks a lot for the bug report.

CJ-Wright:
Sure. I will try out the fix, thank you very much!

vallsv:
Then no backport for 0.12. A 0.13 should be released soon.
I try the dev version again today, and it looks to work.

PR #373, opened 2016-07-29
Python 3 compatibility code (fix issue #368)

vallsv:
Few patches to support both Python 2 and 3


vallsv:
Let me know @payno  if it is fine for you

PR #384, opened 2016-08-04
Clean up compatibility code for setuptools (do not merge)

vallsv:
- It is not possible to use setuptools version >23 and <25.1.4 to compile the project with Python 3.5 on Windows
- setuptools 25.1.4 was patched and is now released


vallsv:
Fix on 25.1.4 will be/was removed.
There is an issue somewhere else.

PR #394, opened 2016-08-24
Back to use extension from numpy (close #367)

### LINKED ISSUE #367: Building problem on Appveyor/Python 3.5/setuptools 24.*/25.*
In the linking process, ```/LIBPATH``` are not quoted. As result path containing spaces are interpreted as object files:

* ```/LIBPATH:C:Program Files...``` became:
* ```LINK : fatal error LNK1181: cannot open input file 'Files.obj'```

I open a bug there
- https://github.com/pypa/setuptools/issues/694

But maybe it is also relative to our setup.py, cause ```silx``` build without any problem on the same environment.

I only force to use ```setuptools==23``` on appveyor configuration file.

vallsv:
- setuptools 25.3.0 was patched
- numpy 1.11.2 was patched (not yet released)
- only few intersections between them are not working

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "GPLv3+"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "04/08/2016"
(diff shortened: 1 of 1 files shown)


vallsv:
The issue was relative to Python 3.5 on Windows machine. Let see what appveyor think about it.

PR #397, opened 2016-08-25
Fix issue 322 average with monitor

vallsv:
Improvement of pyFAI-average

**Mostly**
- Create an own module for average
- Refactoring as object oriented ([average.pdf](https://github.com/silx-kit/pyFAI/files/437594/average.pdf))
- Allow to use a monitor from file header (`--monitor-name=counter/bmon`)
- Allow to request computation of multi filters (`--method=mean,max,min,sum,std`)
- Use a shell progress bar for the process

Some Carlotta Giacobbe provide to me 1 experiment data using a monitor: 38 files, 2GB. I generate different output to try to test the monitor, and compare pyFAI versions to avoid regression:
- (a) pyFAI 0.12
- (b) pyFAI 0.13 before this PR
- (c) pyFAI 0.13 with this branch

**As result**
- With monitor, the result is difficult to evaluate
- Without monitor
    - I notice difference between (a) and (b). I think there is a bug with sum filter in 0.12 (which compute a mean). There is also a changes in filter sum-cutoff, that i can't explain.
    - I do not notice differences between (b) and (c).
- Scripts and data used are available

**Other notes**
- Maybe rework darkAverageFilter (merge the function inside the class)
- We can create an HDF5 output embedding multi filters result (maybe another PR)
- I dont try to generate the documentation

**Todo**

- [x] Store monitor name used in the result header
- [x] Check if files used are stored in result header
- [x] Check API documentation
- [x] Rename utils/string into something else
- [x] Check pixel size of stack filter
- [x] Move `utils.removeSaturatedPixel` into `average` module


vallsv:
Todo:

- [ ] Store monitor name used in the result header
- [ ] Check if files used are stored in result header

vallsv:
All the todos are fixed now.

- I still ask if we can move `utils.removeSaturatedPixel` into average, cause it is not used outside average module.
- I also should check the documentation if there is something about average

kif:
[ ] doc of pyFAI-average --help to explain how to select monitor in counters
[ ] take care of performing operations per frame in 64 bits, especially monitor division

vallsv:
Looks fine

kif:
looks good to me

vallsv:
Sorry, i missing the documentation. Will be part of another commit

PR #407, opened 2016-09-02
Issue 405

kif:
for review ....


kif:
apparently the bug is fixed ... checked on D2AM with WOS + D5

PR #421, opened 2016-09-16
Fix global on RELEASE_LEVEL_VALUE

vallsv:
Reveled when using `./bootstrap.py ./version.py` from make_man

No idea why it was not spotted before.


kif:
on-day it needs to be investigated ...

PR #422, opened 2016-09-16
Convenient PATH fix to generate man files from local project and not from installed scripts

vallsv:
- The inconvenient is it uses hard coded path
- But this script is not often used

Well. I think it is convenient. Cause i would like, at the end to remove all auto-generated content.
But i understand it is not merged, cause it is not portable.


kif:
not the nicest solution but it does the job

PR #428, opened 2016-09-20
Common preproc

kif:
This exposes a preprocessing, both at the C and Python level. It may be used by all Cython based rebinning engines.

Additional features are:
-----------------------------

* Nan are filtered out at input and replaced by dummy/empty
* The flatfield image is filtered out for dummy values (mask from flat)
* Dark and Polarisation correction are unchanged
* There is a new field for coping with sensor efficiency 
* SolidAngle is still present for compatibility reason but should be phased out in a near future


vallsv:
That's ready to merge?

kif:
I guess ... well it is up to you.
there will be another PR when the different classes will use it.

vallsv:
Then it looks fine

PR #432, opened 2016-09-22
fix the Xpad_flat (3.2) detector geometry

kif:
waiting for some test data ...


kif:
I tested the code and the calibration obtained with the new detector definition is indeed better (half residual error).
OK

PR #435, opened 2016-09-27
434 assert

kif:
should be straight forward ... all based on search&replace


vallsv:
There is a typo on `solidangle.size` instead of `solidAngle.size`, 8 or 9 times. Else it looks fine.

kif:
ready to be merged

vallsv:
Looks fine

PR #444, opened 2016-10-17
Bayes

kif:
This brings a bayesian estimator of the background for 1d data (possibly 2d images in the future)


vallsv:
Looks fine

PR #447, opened 2016-10-20
Common preproc

kif:
Start integrating preprocess (OpenMP-parallel, factorized) into:
* pixelwise worker
* distortion correction worker
* correct a bug in AzimuthalInegrator regarding normalization of data


vallsv:
Looks fine

PR #452, opened 2016-10-26
Update GUI with merged silx gui

vallsv:
Merge Qt gui initialization from silx
- there is a new `gui` module with `qt` and `matplotlib` (copy-paste from silx)
- `gui_units` -> `gui.utils`, and only contains few extra functions
- `third_party/pyside_dynamic.py` was moved to `gui/_pyside_dynamic.py` then it is still here when using debian (on debian packages `third_party` is removed)
- there is a simple tests for all scripts, launching them with `--help` (and skipped if qt is not there)
- I should support PySide, PyQt4, and PyQt5. <s>But maybe there is an issue on PySide+Qt5 with matplotlib</s> (according to Thomas PySide is only for Qt4, then it's fine)
- i mostly tested `pyFAI-integrate` and `pyFAI-calib` but everything should be deeply manually tested


kif:
LGTM ... but neither intensively tested

PR #454, opened 2016-10-27
Debian9

kif:
allows the debian packaging to work on debian9. Not yet validated it does not break debian 8 :/


vallsv:
Everything is fine, but it is needed to not remove `StreamHandlerUnittestReady` if you want to have a working `display_buffer` feature

kif:
It is bit dangerous to code too much with meld: I backported some features from silx, shame on me.

vallsv:
Outch

PR #456, opened 2016-10-28
Typo on CCD calibration tutorial

vallsv:


kif:
Hi Valentin,

Thanks for the check-spelling. This is a generated file from jupyter, exported to RST. It would be better to correct the source file.

Cheers, J

vallsv:
Here is an update of the notebook

kif:
LGTM. Thanks a lot for the corrections.

PR #466, opened 2016-11-04
Refactoring benchmark

vallsv:
Mostly
- Typo
- Remove eval and exec
- Remove __name__==__main__
- Relative import of pyFAI lib
- Fix loop between available ocl devices and requested ocl devices


kif:
I merge it  and will test on the mac soon

PR #467, opened 2016-11-07
Issue 462

kif:
this should be tested under windows and linux, python3.5: I dont have the computers here.


vallsv:
I fixed the conflict here https://github.com/silx-kit/pyFAI/pull/468

PR #469, opened 2016-11-08
Issue 459

kif:
actually only this file was impacted


vallsv:
I merge that, but i would be better to clean up your branch. There is a lot of commit noise.

PR #472, opened 2016-11-08
Fix openmp setup

vallsv:
- Change the order of the check of openmp and cython (argument first, then environment)
- add --force-cythonize
- Force cythonize if OpenMP is not expected (assert that default cythonized files was generated with openmp (maybe we should change this?))
- Plus some clean up

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -151,7 +151,7 @@ def check_openmp():
     return True
 
 
-USE_OPENMP = "openmp" if check_openmp() else ""
+USE_OPENMP = check_openmp()
 USE_CYTHON = check_cython()
 if USE_CYTHON:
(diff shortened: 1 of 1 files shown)


CJ-Wright:
Would it be possible to add an OSX dependent line which essentially removes `histogram.c`? This would mean that the user does not need to `rm pyFAI/ext/histogram.c` by themselves.

vallsv:
This patch already force cythonizing everything, then histogram.c will be overwritten in OSX.

Darwin -> force disabling OpenMP -> which force re-generating c files from cython

CJ-Wright:
Awesome! The `histogram.c` issue is/was causing nightmares with conda packaging the project for conda-forge.

PR #473, opened 2016-11-08
enforce the cythonization on OSX ...

kif:
I am not yet sure it is a good idea ...

ideas are welcome


vallsv:
My PR use the `force` param of cytonize. I think it is better. It come from silx.

kif:
You are probably right

PR #478, opened 2016-11-15
add mask and flat to MultiGeometry

picca:
now it should work


kif:
Looks good to me. thanks for the contribution.

PR #487, opened 2016-11-25
Fix geometry documentation

vallsv:
- move  `doc/geometry` into  `doc/mathematica` and remove the tex file
- import the mathematica file into geometry.py
- other clean up of geometry.py documentation (especially using .. math::)


kif:
LGTM

PR #494, opened 2016-11-28
Bunch of fixes

vallsv:
It is better to check it commit by commit

- Fix deadlock in hdf5writer
- Remove some sys.exit
- Fix python 3 syntax in exception
- Fix a lot of warning in unittests (a big commit but nothing very interesting)
- Fix radial range when using pyFAI-integrate
- Normalize python headers


kif:
LGTM

PR #495, opened 2016-11-28
Benchmark

kif:
few typos in benchmarking


vallsv:
Just a comment, ``ocl_devices = []`` have to be defined before ``if ocl:``, else the loop at the and will not work if ocl is dot defined

PR #508, opened 2016-12-07
Non-integer indexing raises error in Numpy >=1.12

woutdenolf:


vallsv:
Thanks Wout. After review we think about fixing `self.massif.find_peaks` instead of this function. Which mean fixing matplotlib events. It will be fixed this week for sure. Thanks a lot for the pull request anyway.

vallsv:
Here is the patch https://github.com/silx-kit/pyFAI/pull/510

PR #510, opened 2016-12-12
Fix matplotlib mouse event

vallsv:


kif:
LGTM. I guess you checked the need for a +0.5.

PR #516, opened 2016-12-14
Issue 515

### LINKED ISSUE #515: [pyFAI-integrate] not using OpenCL in "--no-gui" mode ...
Even if the json file contains:
"do_OpenCL": true

This is a bug

kif:
close #515: use OpenCL when requested, fall back on OpenMP based CSR by default.


vallsv:
I update  `integrate_widget.py`, let me know if it is fine. You can merge it.

PR #524, opened 2017-01-13
Close #523

### LINKED ISSUE #523: Pilatus dx/dy files are not improving the fit
actually it looks worse :(

kif:
Comes from the refinement (pyFAI-calib) of very precise data from ID15/31


vallsv:
Looks good

PR #528, opened 2017-01-30
Backport from 0.14-dev to 0.13

vallsv:
Backport critical bug fix from 0.14-dev
- Convert matplotlib mouse input into pixel coordinates (fix issue with numpy 0.12)
- Fix eiger-mask script if h5py is not installed
- Use "/usr/bin/env python" for scripts


kif:
LGTM. thanks

Issue #526, opened 2017-01-27
New ABC for all OpenCL processing


kif:
This has been validated with ocl-preprocess. 
It should be extended at least to ocl_azim_csr and ocl_azim_lut

kif:
also move the ocl_*.py modules to opencl/* à la silx

PR #532, opened 2017-02-02
Move OpenCL related source file into a sub-module.

kif:
Yet another very big pull-request ongoing ... 

Some tests may be skipped, some files may need rework but looks
promising


kif:
close #526

vallsv:
Usefull refactoring. Looks good.

PR #537, opened 2017-02-09
2d median

kif:
This is not of direct use for pyFAI and may be merged into silx. It is a nice demonstration of usage of the openCL framework


kif:
see code in silx

PR #539, opened 2017-02-10
533 tutorial

### LINKED ISSUE #533: check distortion tutorial ...
Reported by D2AM

kif:
This pull request fixes an actual bug in the distortion correction where a value was not initialized and crashing randomly. Just a typo.

I regenerated 2 tutorials with the current version of jupyter which inlines the images in the rst file. Not a big deal, so I removed images from previous notebooks. The only side effect I see is that those images do not show-up in the latex-generated document from Sphinx. To be discussed with the silx team.

close #533


vallsv:
LGTM

PR #542, opened 2017-02-14
Goniometer and its calibration

kif:
Close a couple of issues related to 
* peak-picker (more useable from jupyter console)
* guess_poni based on ellipses

New test on the geometry on synthetic data.


kif:
I believe it is mergeable as such (if all tests are passing)

vallsv:
BTW would it be possible to create 2 pull requests? One for the peak picker and one for the goniometer.
This PR become very big and i would like to test the picker with ellipsis on my own branches.

kif:
@vallsv 
I think the API is now pretty usable and should be rather stable. Could you have a look and comment ?

kif:
Is it OK now as the test and the tutorials are updated

kif:
Can we merge this ?

vallsv:
Ill take a look.

vallsv:
I send you a PR with a bunch of typo. I also fix the doc.
I still have to check the architecture.

vallsv:
It is non-blocking but here is commands about the python code:
- `readRingNrFromKeyboard` should be removed from `ControlPoints`.
- `Geometry.set_param` is imho not a good idea. I guess there is setters for all the parameters, it takes more lines of code but it is safer. If that's only an util function to the `GeometryRefinment` is should be a private function of this class (that's its responsibility).
- Do you really need to manage `GeometryTranslation = GeometryTransformation`. The code is not yet published, right?
- You can serialize any `trans_function` implementing `to_dict` but you only can deserialize `GeometryTransformation`. Is that dangerous?
- IMHO the relation between `GoniometerRefinement` and `Goniometer` should be a composition and not an inheritance. I know it is the same for `Geometry/GeometryRefinement`. But i dont think it is a good pattern for both (`Geometry` and `Goniometer`). One is a process (`Refinement`) the other one is a state. One is applied to the other one.

I still have to review the notebooks

PR #545, opened 2017-03-06
Improve calibrant api

vallsv:
Here is fixes as discussed:

- Typo on CalibrantFactory class
- Deprecated getitem
- CALIBRANT_FACTORY instead of ALL_CALIBRANTS
- Add eq, ne, hash, copy to Calibrant


kif:
overall looks good to me.

We will have to update the tutorials accordingly.

PR #560, opened 2017-04-20
557 inpainting

kif:
Idea: perform inpainting on the 2D regrouped image to 
* paint less pixels
* take benefit of the quasi-polar symmetry.
then project back to the image geometry using bilinear interpolation.

The final step is to add noise on the inpaintied pixels using a Poissonian distribution.

Nota: this is not science, this is Photoshop.


vallsv:
LGTM. Ill wait for the CI.

vallsv:
Tests fail 2 times on Appveyor. You should investigate.

kif:
Yep, looks like a seg-fault under windows. More work is needed !

kif:
should be OK now...

vallsv:
LGTM

PR #571, opened 2017-05-17
Update setup.py with silx one

vallsv:
- Uses `setup.py` from `silx`
- There is still part of custom code
- [ ] Convert scripts into entry-point (it will change the order of the import then we have to talk about it)
- [ ] Improve the BuildMan target to gen man files from all the entry-point description


kif:
LGTM

PR #572, opened 2017-05-18
Use entry-points instread of scripts

vallsv:
- Update bootstrap from silx
- Use entry-points
    - It removes a lot of scripts and computation from `setup.py`
    - It creates a package `pyFAI.app` for all the entry-points
    - But it change the order of the import cause now application are part of pyFAI
    - Then `pyFAI.__init__` is executed before `pyFAI.app.__init__` which is executed before `pyFAI.app.average`
- Update setup.py `BuildMan` to generate man for all entry-points
- Remove man/pdf files
- Update Debian 8/9 packaging
- I remove `python-tk` from dependencies cause it was causing problem on scisoft12 and it looks to be unused.


kif:
LGTM

PR #581, opened 2017-05-30
Small fixes on the GUI prototype

vallsv:
Small things, but quite useful. Tested with real acquisition images.


kif:
LGTM ... I will test.

PR #602, opened 2017-06-13
587 metadata

### LINKED ISSUE #587: [io] Propagate input metadata into the integrated result
From ID11 feedback at the PyFAI workshop

kif:
close #587


vallsv:
- I don't like the idea of using `res._set_metadata(...)` outside of the integrator as this object is supposed to be non-mutable. `integrate_widget` and `worker` patches the result in this way. It would be better to provide the metadata to the integrate function (this patch is supposed to work like that). Or rework the writer to support metadata.
- On `integrate_widget`, `res._set_metadata(frame.header)` is not called in case of `integrate2d`
- This patch breaks the azimuthalIntegrator class if somebody have overwritten `azimuthalIntegrator.save1D` or `azimuthalIntegrator.save2D`. This functions are not anymore called from integrate1d/2d. If it was known, it's fine for me.

vallsv:
- BTW i would be very interesting if we can provide the EDF metadata used by the file format (black listed ones) from `FabIO`. I would be able to use it from silx. Let me know if it makes sense.

vallsv:
- In `fully_qualified_name(self.ai)`, i think `self.ai` is not defined
- I am not very happy with the `fully_qualified_name`. This pattern is definitively very fragile, it can hide big issues. Then at least we should create unittests.
- Maybe we can provide a method from `azimuthalIntegrator` (or any other classes) to provide specific metadata, to avoid specific code for each integrators in `DefaultWriter`. It also can allow to remove the AI from the constructor. We can talk about it.

kif:
Jakub (ID31) found another bug in caching the headers:

headers (used to) contain the normalization factor which varies naturally from frame to frame.
It does not in the previous version (while calculation are properly performed).

kif:
Merged the branch 601_flatfield here.

kif:
I added some more information about the source of the mask/dark/flat in the result object hence in the saved header. Mask is by the way "new"

vallsv:
LGTM

PR #611, opened 2017-06-19
445 container

kif:
This PR defines a common container for sparse matrix initialization called ArrayBuilder. It does not (yet) provide the usage of the class in the different engines but offers good performances: for a 1k x 1M  array the cost is 186ms (104ms for python lists and  >700 for numpy arrays) then the export to CSR or LUT is <7ms.

The usage of this "ArrayBuilder" will occur only in v0.15, resulting in a huge simplification of most regriding engines.


vallsv:
It would be nice to add at least a test coverage, to check if `as_CSR`, and `as_LUT` returns something.

vallsv:
There is a conflict

vallsv:
There is again a conflict

kif:
I did not encounter any issue when merging...

vallsv:
LGTM

PR #614, opened 2017-06-22
provide the median-filtered instead of the mean along a ring.

### LINKED ISSUE #613: Split "separate" to retrieve the median filtered spectum
Request from ID15

kif:
close #613


vallsv:
I will merge it.

But few comments:

- `method_called` looks to be unused.
- `percentile` is imho not a viable solution (in both `medfilt1d` and the result object), as it is only a parameter of a specific reduction, and you said you will implement more reductions.

But i think i don't really see the way you go. I understand you would like to create a single new function from AI to taking care all this new specific reductions at ones, but it looks like you will create a new one for each new reduction. Is that viable?

Then 2 proposals.

## Reduction object in the result object

At least, as it will share the same result object, it would be interesting to create a description class for this reduction (which will not contain the computation but only the way it have to be done). At least the result object can contains a single object explaining the computation done rings.

- `AbstractReduction()`
- `MeanReduction()` (its the default one? i guess, the one used by integrate1d?)
- `MedianReduction()`
- `CentileReduction(percentile=0.5)`
- `QuantileReduction(0.1, 0.9)`
- or anything you will imagine

## Reduction object as param

If you think you can factorize some code of the integration method of the AI, you also can think about providing this reduction object as parameter of the method. If there is thing to factorize, you can. If no, you can delegate to a specialized private function.

- `integrate1d(azimuthal_reduction=CentileReduction(50))` can call medfiler1d if you like
- `integrate1d(azimuthal_reduction=MeanReduction())` is just the normal integration1d
- ...

My fifty cents, i don't really know where you exactly go, then i don't really know if i understand it well. But i am afraid that everything will grow up and will be painful to maintain.

PR #618, opened 2017-06-23
This PR is to correct small unconsitency between python and cython/OpenCL implementation of preprocessing

kif:


kif:
the bug appears only with non masked "0" in the flatfield image.

0 in the flatfield means the pixel has no weight which is equivalent to being masked.

vallsv:
LGTM

PR #621, opened 2017-07-07
Sigma clip

kif:
close 617


vallsv:
LGTM. Kernel configuration looks much better now.

PR #627, opened 2017-07-18
pyFAI-Integrate with monitor

### LINKED ISSUE #446: [pyFAI-integrate] divide by monitor value
similar to what is done on pyFAI-average

vallsv:
- It added tests for for pyFAI-integrate
    - Therefore there is fix here and there for python, and buggy code
- It uses the monitor to fix the normalization factor: `normalization_factor = monitor`. Maybe it is `1/monitor`? I do it fast
- The normalization was stored on the result
- You can use this kind of param `--monitor-name=counter/mon`
- I do not update the GUI. We can provide a text input. But i presume it will not be a nice solution.

Closes #446


kif:
LGTM

PR #654, opened 2017-08-23
Simple re-naming of the function to avoid the collision with zlib

### LINKED ISSUE #649: Segfault with libz
Using azimuthal integration on Ubuntu 16.04.3 LTS results in a seg fault with the the below backtrace, identifying an incompatibility. I tried building newer versions of libz (zlib) from source, but had the same issue. 

GDB stack trace:

#0  0x00007ffff71d1908 in crc32 () from /lib/x86_64-linux-gnu/libz.so.1
#1  0x00007fffcd3a387a in __pyx_pf_5pyFAI_3ext_7fastcrc_crc32 (__pyx_v_data=<optimized out>, __pyx_self=<optimized out>) at pyFAI/ext/fastcrc.c:1504
#2  __pyx_pw_5pyFAI_3ext_7fastcrc_1crc32 (__pyx_self=<optimized out>, __pyx_v_data=0x7fff8c0feb70) at pyFAI/ext/fastcrc.c:1467
...

kif:
problem with cp27mu in manylinux1

close #649


vallsv:
Would it be possible to rename it `pyfai_crc32`? instead of `my_crc32`? I think it would be cleaner.

kif:
why not

vallsv:
LGTM

Issue #655, opened 2017-08-24
Factorize regrid engines creation/locking/destruction in Azimuthal Integrator


kif:
By re-factorizing the way those large objects are stored, one can reduce dramatically the number of bugs related to locking and destruction.
This is directly related to #652 which is a large memory leak.

Issue #652, opened 2017-08-21
[test] faillure during the Debian process


picca:
Hello here the errors reported on my stretch machine whcih has 2Go of memory...

```
Test pyFAI 0.14.0 from /home/picca/Debian/pyfai/pyfai/.pybuild/pythonX.Y_2.7/build/pyFAI
........................................................................................................................................................................................................................................................ssssssss.....................................E.............................F
Stderr:
Pixel splitting desactivated !
...............................E
Stderr:
MemoryError: falling back on forward implementation
.......ssssssss............................................ss
======================================================================
ERROR: test_pos_lut (pyFAI.test.test_distortion.TestHalfCCD)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/picca/Debian/pyfai/pyfai/.pybuild/pythonX.Y_2.7/build/pyFAI/test/test_distortion.py", line 99, in test_pos_lut
    self.dis.calc_LUT(False)
  File "/home/picca/Debian/pyfai/pyfai/.pybuild/pythonX.Y_2.7/build/pyFAI/distortion.py", line 286, in calc_LUT
    self.lut = _distortion.calc_LUT(self.pos, self._shape_out, self.bin_size, max_pixel_size=(self.delta1, self.delta2))
  File "pyFAI/ext/_distortion.pyx", line 514, in pyFAI.ext._distortion.calc_LUT (pyFAI/ext/_distortion.c:12668)
    return numpy.core.records.array(tmp_ary.view(dtype=dtype_lut),
  File "/usr/lib/python2.7/dist-packages/numpy/core/records.py", line 845, in array
    new = new.copy()
MemoryError

======================================================================
ERROR: test_bug170 (pyFAI.test.test_bug_regression.TestBug170)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/picca/Debian/pyfai/pyfai/.pybuild/pythonX.Y_2.7/build/pyFAI/test/test_bug_regression.py", line 89, in test_bug170
    ai.integrate1d(self.data, 2000)
  File "/home/picca/Debian/pyfai/pyfai/.pybuild/pythonX.Y_2.7/build/pyFAI/azimuthalIntegrator.py", line 2550, in integrate1d
    dpos0 = self.array_from_unit(shape, "delta", unit, scale=False)
[...]

PR #659, opened 2017-08-25
655 engines

kif:
close a couple of issues related to large memory consumption in tests:
#655: better handlig of cached engines
#652: skip large tests
+ new Cython generated files (version 0.25.2)


vallsv:
LGTM

PR #661, opened 2017-08-28
Provide the rotation matrix as example in geometry

kif:
This provides a way to retrieve the rotation matrix from a geometry at the python level.

@jonwright could you review the patch ?


kif:
On Thu, 31 Aug 2017 05:14:59 -0700
Jon Wright <notifications@github.com> wrote:

> jonwright approved this pull request.
> 
> Looks good to me ! 
> It seems to be checked via test_XYZ in test_geometry so it should be correct.
> Thanks for making these changes

I added a few of tests on chi as well to ensure the self consistence.

PR #663, opened 2017-08-31
V0.14

kif:
A hack to enforce the name resolution at link time which works on manylinux1


vallsv:
Just 2 comments:
- You should import crc32.h from crc32.c, then you will not have to declare 2 times the HIDDEN defines.
- It would be better to work on the master, will that's not specific to this branch. We can backport expected things to the 0.14.2 when we think it's fine.

vallsv:
There is still a `my_crc32` on the code. I think it would really be better and easy to review to close this PR. And to create a PR on the master features per features. We can see it together if you like.

vallsv:
I close this PR while it is now split into PRs:
- https://github.com/silx-kit/pyFAI/pull/671
- https://github.com/silx-kit/pyFAI/pull/672
- https://github.com/silx-kit/pyFAI/pull/673
- https://github.com/silx-kit/pyFAI/pull/674

PR #677, opened 2017-09-06
Skip OpenCL if requested

### LINKED ISSUE #667: [test] please provide a way to unactive the OpenCL tests
Hello it would be nice if the OpenGL test could be unactivated via an environment variable.

It cause  troubles during the debian package build process.

thanks

Fred

vallsv:
- Can be done by run_tests.py --no-opencl
- Or using envvar PYFAI_OPENCL=False

Closes #667


kif:
LGTM

PR #683, opened 2017-09-08
670 opencl

kif:
fix regression of performances


vallsv:
What about using attributes instead of keys? 
```
self.kernels.mean_std_horizontal(self.queue, ws, (1, wg), *kargs.values())
```
Instead of
```
self.kernels["mean_std_horizontal"](self.queue, ws, (1, wg), *kargs.values())
```
The result will be much easy to read and use, while it is callables.

I can do the job if you like.

vallsv:
LGTM

vallsv:
There is no way to have constant while everything come from opencl program.
It have to be generic, cause it is a generic loader of opencl kernel.

> whcih can not be catched until executed

Welcome to the Python world ;-)

PR #703, opened 2017-10-03
[calib2] Use the fitted wavelength to save the poni file

### LINKED ISSUE #694: pyFAI-calib2 saves wrong poni if the wavelength has been fitted
In pyFAI-calib2 you unlock and fit the wavelength, then save the poni file. Result: the wavelength in the poni has the starting value, but not the fitted one.
It also seems that both the cake and the integrated image are also done using the starting value of the wavelength, but it is just a feeling which is not easy to prove.

vallsv:
Closes #694


kif:
You expose wavelength in Angstrom,
Would it make sense to expose angles in degrees  in the same way ?

PR #706, opened 2017-10-09
Gonio wavelength

kif:
This PR allows to fit the wavelength in a goniometer definition function.
It resolves also a long lasting bug when the wavelength was increasing and the number of rings was shrinking. This PR should limit the effect.


vallsv:
LGTM

PR #708, opened 2017-10-10
Fix jupyter rendering

### LINKED ISSUE #707: [multicalib] ring with zero points
Hello, I did an `sg.extract_cp(pts_per_deg=pts_per_deg)` and for one of my images, I got this result

```
scan_3_01_10
ControlPoints instance containing 19 group of point:
LaB6 Calibrant with 245 reflections at wavelength 4.85945727522e-11
Containing 19 groups of points:
#ek ring 1: 36 points
#el ring 2: 70 points
#em ring 3: 72 points
#en ring 4: 75 points
#eo ring 5: 77 points
#ep ring 6: 8 points
#eq ring 7: 79 points
#er ring 8: 81 points
#es ring 9: 65 points
#et ring 10: 31 points
#eu ring 11: 16 points
#ev ring 12: 43 points
#ew ring 13: 30 points
#ex ring 14: 24 points
#ey ring 15: 18 points
#ez ring 16: 13 points
#fa ring 17: 8 points
#fb ring 18: 4 points
#fc ring 19: 0 points
```

as you can see, the last ring is empty.

when I try to display the image with `jupyter.display(sg=sg)`, I got this error message

```---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-106-1b1b900f5a13> in <module>()
      1 # Initial refinement of the translation table model
      2 
----> 3 optimize_with_new_images(multicalib, gonioref)

<ipython-input-105-302f714dc06c> in optimize_with_new_images(multicalib, gonioref, pts_per_deg)
     21     if sg:
     22         sg.geometry_refinement.set_param(gonioref.get_ai(sg.get_position()).param)
---> 23         jupyter.display(sg=sg)

/usr/lib/python3/dist-packages/pyFAI/gui/jupyter.py in display(img, cp, ai, label, sg, ax)
     74         for lbl in cp.get_labels():
     75             pt = numpy.array(cp.get(lbl=lbl).points)
---> 76             ax.scatter(pt[:, 1], pt[:, 0], label=lbl)
     77         if ai is not None and cp.calibrant is not None:
     78             tth = cp.calibrant.get_2th()

IndexError: too many indices for array
```

so it seems to me that a ring with zero points should not be removed :)

Cheers

Fred

vallsv:
Closes #707

### DIFF
--- a/pyFAI/gui/jupyter.py
+++ b/pyFAI/gui/jupyter.py
@@ -35,7 +35,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "29/05/2017"
+__date__ = "10/10/2017"
 __status__ = "Production"
 __docformat__ = 'restructuredtext'
 
@@ -84,7 +84,7 @@ def display(img=None, cp=None, ai=None, label=None, sg=None, ax=None):
 
 def plot1d(result, calibrant=None, label=None, ax=None):
     """Display the powder diffraction pattern in the jupyter notebook
-    
+
     :param result: instance of Integrate1dResult
     :param calibrant: Calibrant instance to overlay diffraction lines
     :param label: (str) name of the curve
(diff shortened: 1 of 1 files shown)


picca:
wouldn't be better to avoid creating a ring with zero points ?

picca:
in the `geometry.py` the `extract_cp` function contain this

```
               res = self.massif.peaks_from_area(mask2, Imin=0, keep=keep, dmin=dist_min, seed=seeds, ring=i)
                cp.append(res, i)
```

maybe res should return None when nothing was found and c pshould be modify only if res != None

vallsv:
The jupyter rendering have to fixed anyway. I have no idea with the other part.

PR #716, opened 2017-10-26
Packaging ubuntu

kif:
minor fix for the packaging


vallsv:
Just a note, lines starting with `target_system` looks to be missaligned.

vallsv:
LGTM

vallsv:
I still have problem, help2man is not needed to have a working `python3-pyfai`, then there is something strange in this dependencies.

PR #718, opened 2017-10-27
715 azimuthal variance

### LINKED ISSUE #715: Add non regression test on "azimuthal" error model (pyFAI.AzimuthalIntagrator.integrate1d)
reported by Marco Cammarata

kif:
close #715


vallsv:
`memory=None` is not used, it that normal? Else it looks fine.

kif:
On Tue, 31 Oct 2017 01:00:26 -0700

> `memory=None` is not used, it that normal? Else it looks fine.

this keyword option was added for consitency with
silx: it is used in sift and we aim at using the one from silx. 
until then, it is better to have the same signature for the functions
to prepare drop in replacement.

Cheers

-- 
Jérôme Kieffer
tel +33 476 882 445

PR #727, opened 2017-11-24
Fix ring number for input/output to npt

vallsv:
Let me know if it works fine.


kif:
@dnaudet Can you check if it work for you ?

PR #734, opened 2017-12-04
user h5py for read_the_docs installation

kif:
RTD fails in building the documentation recently ... there is an error mesage about fabio.nexus. h5py may be missing


kif:
I will switch the doc generation to nbsphinx ... no more intermediate rst file

vallsv:
LGTM

PR #744, opened 2017-12-18
Fix inpainting typo

### LINKED ISSUE #739: [Inpainting] maybe fixed upstream
> I'm making a simple script to use the inpainting ad for some reason the
> inpainting() method gets stuck (run on rnice8) without any errors.
> 
> python /data/id31/inhouse/scripts/pyFAIWInpaint.py -j .azimint.json -o .
> empty_spin_P3_0001_0000.edf
> 
> run from /data/visitor/ma3624/id31/Agnieszka folder.
> 
> The problem is on line 235.
> 
> When I use it in the jupyter notebook (
> http://scisoft13:8000/user/drnec/notebooks/temp/Inpainting.ipynb), it is
> perfectly fine.

vallsv:
According to the data, the image was fully masked, then no valid pixels was available.

This PR:
- Avoid infinite loop
- Provide a runtime exception explaining there is no valid pixels

Close #739

### DIFF
--- a/pyFAI/azimuthalIntegrator.py
+++ b/pyFAI/azimuthalIntegrator.py
@@ -32,7 +32,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "27/10/2017"
+__date__ = "20/12/2017"
 __status__ = "stable"
 __docformat__ = 'restructuredtext'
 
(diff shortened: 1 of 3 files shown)


vallsv:
I have to commit the pre-compiled cython files

kif:
Just a comment: I would advise to comment the "assert" statement used in cython files as neither the source code (.pyx), not the transcoded C file are present on the computer running the "assert" statement

kif:
I leave you update the Cython files.

vallsv:
Maybe it was an old behaviour cause this code without comment was raising assertions in release mode (using the bootstrap). Then it is working.

vallsv:
Ah, you mean to provide more information to the assert? Then yes, that's a good idea.

vallsv:
Looks to be good now

kif:
LGTM
close #739

PR #759, opened 2018-01-22
Issue url

kif:
Fix issue reported by Natalie Geise when the image filename is specified as c:\bla\img.tiff which should not be interpreted as an URL.

There are a couple of backports of FabIO's last packaging sequence as well.


vallsv:
LGTM

PR #769, opened 2018-01-25
Regression 758

### LINKED ISSUE #758: Update constants value
Use the one from scipy to avoid having to maintain physical constants:
`pyFAI.units.hc = scipy.constants.c*scipy.constants.h/scipy.constants.e*1e5`

kif:
to prevent further regression: ensure 6 dgit precision
close #758


kif:
oups. wrong stating point

PR #770, opened 2018-01-25
non-regression test on hc constant

### LINKED ISSUE #758: Update constants value

kif:
close #758


vallsv:
LGTM

PR #776, opened 2018-02-01
775 calibrant

### LINKED ISSUE #775: issue in calibrant
```
In [13]: l=pyFAI.calibrant.get_calibrant("LaB6")
    ...: l.wavelength=1e-10
    ...: l.get_2th_index(0.1)
    ...: 
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/users/kieffer/workspace-400/pyFAI/bootstrap.py in <module>()
      1 l=pyFAI.calibrant.get_calibrant("LaB6")
      2 l.wavelength=1e-10
----> 3 l.get_2th_index(0.1)

/users/kieffer/workspace-400/pyFAI/build/lib.linux-x86_64-2.7/pyFAI/calibrant.pyc in get_2th_index(self, angle)
    533         idx = None
    534         if angle:
--> 535             idx = self._2th.find(angle)
    536         if idx == -1:
    537             idx = None

AttributeError: 'list' object has no attribute 'find'
```

kif:
close #775 
offer a selection of the precision in addition

### DIFF
--- a/pyFAI/calibrant.py
+++ b/pyFAI/calibrant.py
@@ -41,7 +41,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "01/02/2018"
+__date__ = "02/02/2018"
 __status__ = "production"
 
 
@@ -384,7 +384,7 @@ class Calibrant(object):
         :rtype: Calibrant
         """
         return Calibrant(filename=self._filename,
-                         dSpacing=self.dSpacing,
+                         dSpacing=self._dSpacing,
                          wavelength=self._wavelength)
 
     def __repr__(self):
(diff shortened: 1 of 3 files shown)


kif:
On Thu, 01 Feb 2018 16:03:18 +0000 (UTC)
picca <notifications@github.com> wrote:

> picca commented on this pull request.
> 
> In my case angle is negatif. I am wondering if we need to take tis into account, or if the user must deal with this by himself.

2theta cannot be negative ... can it ?

picca:
Indeed the theoretical tth are > 0. But rela motors of diffractometer can have negativ values.

In my case, the 2C diffractometer of CRISTAL, tth2C > 0 when the diffraction goes to the roof and < 0 when it goes down. And it is true, that during scan I can reconstruct a scan which goes from negativ values to positif values.
After this I need to fold in order to projet the negativ part into the positiv one.

Cheers

kif:
Apparently it breaks some tests ... 
test_2th in pyFAI.test.test_calibrant.TestCalibrant (l76)
"We expect 25 rings for LaB6" some more work is needed.

For the signed version of tth, I don't agree. If you goniometer is mounted offseted, you should correct the value before and not expect pyFAI to guess it for you.

vallsv:
Looks to be fine then let's merge it

PR #795, opened 2018-02-26
Refactoring for pyFAI 0.16

### LINKED ISSUE #302: Force test coverage to provide indicator from from the full base code
The test coverage is only based on tested files.

### LINKED ISSUE #306: Clean up .gitignore
Take care of root/deep directory to ignore.

### LINKED ISSUE #344: Load resources using resources module

### LINKED ISSUE #651: [pyFAI] Use silx.gui.qt instead of pyFAI.gui.qt
In order to avoid multiple (cut and past), should you use the one from silx now that pyFAI depends on silx.

This is DRY ;)

Cheers

### LINKED ISSUE #746: Add debug logs for try/except imports

### LINKED ISSUE #788: use pyFAI without opencl and avoid annoying warnings
We're using pyFAI, but not using opencl.  Every time we run a program that imports pyFAI, we get a Warning of 

    WARNING:pyFAI.opencl.common:Unable to import pyOpenCl. Please install it from: http://pypi.python.org/pypi/pyopencl

Well, we are not planning on using opencl for every use of pyFAI., and it would add a pretty big dependency  to our code.  

We can set the environment variable `PYFAI_OPENCL' to '0' to change this message to 

    WARNING:pyFAI.opencl.common:Use of OpenCL has been disables from environment variable: SILX_OPENCL=0

(seems like a misleading message, BTW).  But we don't see a way to turn such messages off.   

Is it possible to use pyFAI with this message being written every time?   Perhaps we could set environmental variable of 'PYFAI_OPENCL'  to 'WE_KNOW_IT_IS_MISSING' or something similar to suppress these warnings?

Thanks!

### LINKED ISSUE #793: Fix silx deprecation warnings
For example
```
WARNING:silx.DEPRECATION:module /home/valls/.local/lib/python2.7/site-packages/silx/gui/plot/PlotActions.pyc is deprecated since silx version 0.6. Reason: PlotActions refactoring. Use 'plot.actions' instead.
  File "/home/valls/.local/lib/python2.7/site-packages/silx/gui/plot/PlotActions.py", line 38, in <module>
    since_version='0.6')
```

vallsv:
Closes #746 
Closes #306 
Closes #651 
Closes #344 
Closes #793 
Closes #788 
Closes #302


pyFAI/detectors/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+ALL_DETECTORS = Detector.registry
+detector_factory = Detector.factory
+load = NexusDetector.sload
+
+# Expose all the classes, else it is not part of the documentation
+import inspect
+__all = locals().items()
+__all = filter(lambda i: inspect.isclass(i[1]), __all)
```
kif: I would prefer a list comprehension instead of functional programming (filter and anonymous functions)

vallsv:
PR reworked as requested.

kif:
LGTM

PR #796, opened 2018-03-02
Cookbook

kif:
Provide 2 simple cookbooks for azimuthal integration


vallsv:
It would be good to remove the tooltip displayed over the screenshot, else it's fine.

PR #802, opened 2018-03-07
close #801

### LINKED ISSUE #801: No gonio calibration when every image has only 1 peak
This is a typo in the goniometer.py

kif:
Fred (@picca) & Erik from soleil will be happy


vallsv:
LGTM

PR #809, opened 2018-03-15
Bug peak picker

kif:
Beside the move of the import, I added a test that tries to import every single module.

It spotted a number of bug & regression already existing which have been fixed.

Let's see if it passes everywhere.

The last commit is related to another bug, but nothing special there.


kif:
I have to admit the additional test is fragile

vallsv:
LGTM

I was sur `pyFAI_crc32` was already patched in this way...

And nice fixes on many imports.

PR #818, opened 2018-04-05
816 bug hist2d

### LINKED ISSUE #816: Check reset azimuthal angles in geometry ...
Reported by Andreas ID13.

It would be interesting to have a reset_azimuthal and to test for regression as this is a nasty caching bug.

kif:
This time, we actually check the chi-array is properly discontinued at the proper place
close #816

### DIFF
--- a/pyFAI/__init__.py
+++ b/pyFAI/__init__.py
@@ -32,7 +32,7 @@ from __future__ import absolute_import, print_function, with_statement, division
 
 __author__ = "Jérôme Kieffer"
 __license__ = "MIT"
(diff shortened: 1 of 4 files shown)


vallsv:
LGTM

vallsv:
BTW in `geometry.py` i dont understand why you import `pi` from `numpy`, and you use it as `numpy.pi`. It looks like most of this import line is useless.

```from numpy import radians, degrees, arccos, arctan2, sin, cos, sqrt, pi ```

IMO it is always better yo use this as `numpy.something`, but it is up to you.

PR #819, opened 2018-04-10
Generalize cached array

kif:
This generalized the usage of cached array for:

* the cosine of the incidence angle
* the solid_angle array
* the transmission correction array

The code is simpler and less error prone. 
Properties are providing compatibility with former API


vallsv:
LGTM

PR #836, opened 2018-05-03
831 id002 variance

kif:


kif:
This addresses for now only pixel-wise worker ... but is may be better to merge smaller parts.

PR #844, opened 2018-06-07
Minor clean up in benchmarking

kif:
* fix exception when no display
* smoother and larger Y axis by default


vallsv:
LGTM, i only revert the typo on the negative numbers.

PR #846, opened 2018-06-08
WIP: add perkin 4k detector

CJ-Wright:
This produces support for the new Perkin Elmer 4k detectors
@jrmlhermitte

### DIFF
--- a/pyFAI/detectors/_others.py
+++ b/pyFAI/detectors/_others.py
@@ -155,7 +155,7 @@ class Perkin4K(Detector):
     DEFAULT_PIXEL1 = DEFAULT_PIXEL2 = 100e-6
 
     def __init__(self, pixel1=100e-6, pixel2=100e-6):
-        super(Perkin, self).__init__(pixel1=pixel1, pixel2=pixel2)
+        super(Perkin4K, self).__init__(pixel1=pixel1, pixel2=pixel2)
         if (pixel1 != self.DEFAULT_PIXEL1) or (pixel2 != self.DEFAULT_PIXEL2):
             self._binning = (int(2 * pixel1 / self.DEFAULT_PIXEL1), int(2 * pixel2 / self.DEFAULT_PIXEL2))
             self.shape = tuple(s // b for s, b in zip(self.max_shape, self._binning))
--- a/pyFAI/detectors/_others.py
+++ b/pyFAI/detectors/_others.py
@@ -144,6 +144,30 @@ class Perkin(Detector):
             (self.name, self._pixel1, self._pixel2)
 
 
+class Perkin4K(Detector):
+    """
+    Perkin detector
+
+    """
+    aliases = ["Perkin4K detector", "Perkin Elmer 4K"]
+    force_pixel = True
+    MAX_SHAPE = (4096, 4096)
+    DEFAULT_PIXEL1 = DEFAULT_PIXEL2 = 100e-6
+
+    def __init__(self, pixel1=100e-6, pixel2=100e-6):
+        super(Perkin, self).__init__(pixel1=pixel1, pixel2=pixel2)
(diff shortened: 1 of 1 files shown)


CJ-Wright:
I think this is ready for review!

jrmlhermitte:
I asked Li Ji. The settings look good.
The detector can be binned but when it is not the dimensions are as written and pixel size as written:
http://www.perkinelmer.com/PDFs/downloads/011104A_01-PRD.pdf

CJ-Wright:
The pixel size is also smaller, at 100 um a side rather than 200 um.

kif:
On Sun, 10 Jun 2018 06:04:46 -0700
"Christopher J. Wright" <notifications@github.com> wrote:

> The pixel size is also smaller, at 100 um a side rather than 200 um.

So we agree, the detector definition in pyFAI is enougt as it already
supports this 100µm & 4096 pixel detector from Perkin. The default
binning is just different.

In [1]: import pyFAI

In [2]: p = pyFAI.detector_factory("Perkin")

In [3]: print(p)
Detector Perkin detector	 PixelSize= 2.000e-04, 2.000e-04 m

In [4]: print(p.binning)
(2, 2)

In [5]: print(p.shape)
(2048, 2048)

In [7]: p.set_binning((1,1))

In [8]: print(p)
Detector Perkin detector	 PixelSize= 1.000e-04, 1.000e-04 m

In [9]: print(p.binning)
(1, 1)

In [10]: print(p.shape)
(4096, 4096)

In [11]: another = pyFAI.detector_factory("Perkin")

In [12]: another.guess_binning((4096,4096))
Out[12]: True

In [13]: print(another, another.binning, another.shape)
(Detector Perkin detector	 PixelSize= 1.000e-04, 1.000e-04 m, (1, 1), (4096, 4096))

In [14]: third = pyFAI.detectors.Perkin(1e-4, 1e-4)

In [15]: print(third, third.binning, third.shape)
(Detector Perkin detector	 PixelSize= 1.000e-04, 1.000e-04 m,
(1, 1), (4096, 4096))

CJ-Wright:
Is there a string alias for that setting?

kif:
On Sun, 10 Jun 2018 08:55:40 -0700
"Christopher J. Wright" <notifications@github.com> wrote:

> Is there a string alias for that setting?

I don't think so. 

But the guess-binning method is called in all code using the factory.
And for third-party code, they (or you) should take care of that.

Writing one class per binning setting will cause an explosion of the
code base on the one hand and an explosion of the number of (stupid)
bugs on the other. Including the binning in alias is an option, but
again, this will explode the number of classes exposed in the GUI which
is not something desirable.

vallsv:
> Writing one class per binning setting

Isn't it about creating a new detector instead of a new binning?

I would prefer a strict definition of detectors: only the constant things (pixel size, number of pixels), which are constraints. As detectors are designed the binning was not something you supported to set but something computed for you. Then if you have to force a binning only for this specific detector, it's not a good idea IMO.

Plus, it would be much better for the GUI to display the specific Perkin detector if it exists instead of a generic one. We anyway need somewhere a name, then we anyway have to define it somewhere.

vallsv:
Sorry, i misunderstand what Jerome said.

After checking the original detector (`Perkin` class), i can confirm that both are the exact same detector (`Perkin` and `Perkin4K`). The only difference is the default binning value at the construction. Then your `Perkin4K` is not needed.

On side note, I think it would be better to use 100um as the `DEFAULT_PIXEL` which became consistent with the `MAX_SHAPE`. But it will not change the behavior of the detector, cause this class attribute is private.

vallsv:
Here is the dispaly on our calibration GUI. You can see that there is no problem to import an image of 4096x4096 on the default perkin detector (binning=1). I do not display the pixel size, but it could be done.

perkinelmer

Then, is there still a problem on your side?

PR #850, opened 2018-06-15
Make classes pickleable (Calibrant, Geometry, Detector)

ronpandolfi:
This merge will make Calibrant, Geometry, and Detector picklable, preserving state. Lock objects are sanitized from the state before pickling.

This provides support for: 
- saving state via more common serializers 
- serializing for multiprocessing
- workflow-based execution which requires pickle-ish serialization (Dask)


kif:
Hi Ron,
Detector and Geometry should probably be pickled separately ...
Your PR seams interesting and I understand the reason but I am afraid this will be much more complicated as both Detector and Geometry have cached and states (i.e. binning for Detector).

I can help you and/or advise you but not before Tuesday
Cheers,

ronpandolfi:
No problem, take your time. I don't have enough awareness of the internals of pyFAI to know what problems this would cause, so let me know what needs to be done.

I also have some tests I can add to this PR.

kif:
On Sat, 16 Jun 2018 23:37:18 -0700
Ronald Pandolfi <notifications@github.com> wrote:

> No problem, take your time. I don't have enough awareness of the internals of pyFAI to know what problems this would cause, so let me know what needs to be done.

I come back to you tomorrow. 

For you information, the definition of the detector in the constuctor
of Geometry remains (for compatibility) from the era where detector
were not yet a separated class.

This is why I believe it would be interesting to make detectors picklable before geometry.

> I also have some tests I can add to this PR.

This would be great !

CHeers,

vallsv:
Hi. IMHO, it have to be checked accurately.
- The calibrant should not contains `filename` nor `wavelength`. Only the `dSpacing` should be needed.
- `pixel1`, `pixel2`, and `splinefile` could be removed from `Geometry`
- Detector could be tricky as it is not a final class
- The black list is not a good idea IMO. It's easier and more explicit to copy what we need to copy.
- I don't really see why you want to avoid `engines` as it is not part of `Geometry`

I am afraid you want to do it for a very special use case. It is maybe better to implement it outside of the classes. I don't know much pickle, but i guess it is possible.

kif:
Hi Ron,

just a question, do you wish to pickle the "cached" data?
* pro: object are ready to use after unpickling
* con: large pickled volume (~ 10x the size of one image)


Cheers, 

Jerome

ronpandolfi:
Hi Jerome and Valentin, Yes, please review this carefully. I think this is a good starting point from which we can discuss needs and solutions.

Regarding caching data, I would take the option that is best performing. I expected that the cost to pickle/unpickle the cache was less than the cost to recompute it, particularly with multi-processing being the typical use case for pickle support. 

- I cleared `engines` because it had locks; should the engine classes also be made pickleable then?
- I opted for a blacklist over whitelist because I didn't want to interfere with any attributes added by subclasses (if any exist), and the length of attributes to blacklist is relatively short and well defined.

An alternative approach could be to make the locks etc. pickleable, without state, and then all other classes would work without change.

kif:
you said you have some tests ...

ronpandolfi:
Thanks, Jerome and Valentin!

kif:
On Thu, 28 Jun 2018 09:36:58 -0700
Ronald Pandolfi <notifications@github.com> wrote:

> Thanks, Jerome and Valentin! 

I merged without Appveyor checks passing ...

There are bugs probably related to the new RC of numpy, all our
projects are affected.

Cheers,
J.

PR #853, opened 2018-06-19
[calib2] Fix "wrong size" with Frelon detector

### LINKED ISSUE #851: [calib2] "size not valid"
occues when defining "frelon" then spline (2kx2k) then loading image (2kx2k)

investigation needed

vallsv:
There was a missing case in `guess_binning`.

I can't really manage all the cases as it is not very clear to me why for example we couldn't use a shape smaller than max_size in this case. Maybe it is cause of the splinefile?

Closes #851


kif:
LGTM. Thanks

PR #854, opened 2018-06-19
Ocl profile

kif:
This is a simple correction for a couple of inconsistencies in the OpenCL part.
Used for benchmarking


vallsv:
LGTM

vallsv:
It would be better to use a specific `if` instead of a very generic `try/except`, but it's fine as it is.

PR #861, opened 2018-07-03
close #860

### LINKED ISSUE #860: guess_binning in detectors erases the mask.
There are 3 things to do to :
* [ ] Check the various scripts calling guess_binning to ensure the mask is still OK
* [ ] Return in guess_binning if the shape is already the expected one
* [ ] Document this "odd" behaviour of guess_binning

kif:
This is an attempt to address #860


vallsv:
Back to numpy blas problems on Windows.
```
    from scipy.linalg import _fblas
ImportError: DLL load failed: The specified module could not be found.
```

vallsv:
LGTM

PR #863, opened 2018-07-09
Provide detector configuration within the PONI file

kif:
Maybe there is a need for a set_config for detector not compatible with default constructor...


vallsv:
Just 2 comments with spaces in config keys:
```
+                key = words[0].strip().lower().replace(" ", "_")
                 try:
                     value = words[1].strip()
                 except Exception as error:  # IGNORE:W0703:
                     logger.error("Error %s with line: %s", error, line)
                 data[key] = value
+        version = float(data.get("PoniFile version", 1.0))
```
If you put spaces in keys, you should read spaces. Else it will create a lot of mistakes.
Here is one. There is also a problem with upper case here. `PoniFile version` will never exists in your dictionary.

It would be better to use integer for versions. Anyway there is a big or a small change, there is a change anyway then. IMO, make it simple.

vallsv:
LGTM

PR #893, opened 2018-08-13
Remove isocontour from cli calib

vallsv:
- Remove pyFAI isocontour from cli calibration
- Remove pyFAI isocontour from goniometer calibration

It have to be tested strongly. I don't try it.

BTW, using a non square image to test is useful to catch axes problems.

Relative to #881


kif:
It could break in some tutorials ... 
but your code looks OK and even if you have forgotten somewhere to change, it will just trigger some future bugs.

Issue #520, opened 2017-01-04
Next generation averaging mechanism: follow the path


kif:
This issue describes the transition path from the current integration mechanism to the next generation.

Currently:
Preproc: I = (Raw-Dark) / (Flat x SolidAngle x Polarization)
Averaging: Im = histogram(pos, weights=I) / histogrram(pos) 

Next Generation:
PreprocN: In = (Raw-Dark) 
             D: Id = (Flat x SolidAngle x Polarization)
Averaging: Im = histogram(pos, weights=In) / histogram(pos, weights=Id)

Advantage: Mathematically better with tilted detector, allows error propagation, ...
Draw-back: may be twice slower ! but we are fighting to make it fast again !

During this refactoring one will need to store 2 temporary arrays instead of one. I suggest to store it as and "array of struct" to allow subsequently to read it back again as an array of float2 in OpenCL.

During the refactoring, I would like to drop the "LUT" representation in favour of the "CSR" which is almost always faster.

This creates the given path:
- [x] write as pre-processing with split-output (array of struct: (num, denom)) in Cython
- [x] write the same in pure Python and validate
- [x] Write once again the same in OpenCL
- [x] Integration using 2 weighted numpy histograms
- [x] Integration using scipy.sparse matrix multiplication based on the CSR look-up table
- [x] Integration using cython histograms, combined, without splitting. Is it still faster ? 
- [x] Integration using cython sparse matrix multiplication, combined
- [x] Integration using OpenCL sparse matrix multiplication based on float2.

Each step will require a bunch of tests to validate the result.

CJ-Wright:
You might also try Numba to help speed things up: http://numba.pydata.org/

kif:
On Wed, 04 Jan 2017 09:30:58 -0800
"Christopher J. Wright" <notifications@github.com> wrote:

> You might also try Numba to help speed things up: http://numba.pydata.org/

I read a bit about numba ... it is very promising; it looks like the
future as it is just-in time compiled code which can run on multiple type of hardware.

I see a few problems:
* The CPU (LLVM optimized) code does not look parallel.
* The GPU code can run on nvidia hardware using CUDA or AMD hardware
  using HSA but the code needs to be adapted each time which does not
  make it portable.

With PyOpenCL we already got parallel code, compiled as needed and
running efficiently both on CPU and on GPU: the CPU code has been
compared with ahead of time compiled code from Cython and OpenCL still
is slightly faster than C-code. 

[...]

PR #913, opened 2018-09-27
882 error propagation without splitting

### LINKED ISSUE #882: saxs, histogram mode: errors are slightly under-estimated
pyFAI in histogram mode slightly under-estimates the errors. The figure below was compiled from pair-wise comparisons of 1390 consecutive SAXS frames of water obtained at P12 at PETRA-III. The frames are similar up to noise as per cormap test as implemented in datcmp of ATSAS, therefore any deviations from the expected chi^2 distribution should be from the error estimates.

chi2-pval

Panel (a) shows the empirical distribution of chi^2 values of pyFAI in histogram mode, panel (b) the corresponding p-values. Note that the distribution in (a) is not centered on 1.0, but shifted slightly to the right (1.0171) and the distribution in (b) would be expected to be uniform. For reference, panel (c) shows the empirical histogram of chi^2 values of data averaged by radaver of ATSAS; panel (d) shows the corresponding, mostly uniform, p-values.

While the effect on the chi^2 values does not appear to be large in absolute numbers, the effect on the p-value distribution in (b) is severe: 3.3% of the chi^2 values have a p-value of 0.01 or less and 12.2% of the chi^2 values a p-value of 0.05. Expected would be 1% and 5%, respectively. In other words, if one uses the p-value as a decision criterion whether there are differences, the pyFAI results indicate differences where there are none two to three times as often as one would expect by chance. Again, for comparison, radaver has 1.2% and 5.6% significant results at the 1% and 5% levels, respectively.

kif:
This is a PR for enabling a proper error propagation in the case of pixel-spliting is de-activated.
The new integration method is prefixed with an "_" as it may disappear and will change a lot before being settled down.
This is directly linked to #520  where the code is implemented without splitting. 

Last notice, I will update the tutorial which demonstrate how it works but there is unfortunately no tests possible. The tutorial requires gigabytes of RAM and lasts a few minutes as it generates thousands of images and spectra.
close #882


vallsv:
LGTM

PR #916, opened 2018-10-02
Opencl coef power

kif:
This implements OpenCL sparse matrix dense vector multiplication with variance propagation ....


vallsv:
LGTM

PR #922, opened 2018-10-04
Add naive tests

### LINKED ISSUE #917: [calib2] Add some test and test coverage
It could be useful to test common interaction to avoid regression.

As the application is still experimental and should change, it is not a good idea to create a big and complex test suite, but some specific tests could be useful.

vallsv:
Test the GUI using an initialized GUI model.

There is no mouse interaction.

Closes #917


kif:
LGTM

PR #925, opened 2018-10-05
fix the test coverage part in order to release v0.16

kif:
small PR for better test coverage. No files remain unloaded (hence a low coverage)


vallsv:
LGTM

PR #941, opened 2018-10-18
Fix file license

vallsv:


kif:
thanks for spotting it.

PR #945, opened 2018-10-18
Cython coef power

kif:
* Update the coef_power for splitPixel
* Use the common data_types
* fix some debian packaging scripts


vallsv:
LGTM

PR #977, opened 2018-10-30
Fix validators

### LINKED ISSUE #975: [calib2] '.' in the input text cause some troubles.
Hello, when I try to change an entry like the wavelength and press the '.' the cursor doesn not position itself after the dot, but before.

So instead of

1.54

I have

154.

vallsv:
Closes  #975


kif:
Let's see if it helps :)

PR #979, opened 2018-10-30
[calib2] Add range constraints

### LINKED ISSUE #875: [calib2] Add bound constrains on fitting

vallsv:
Allow to set range constraints 

- [x] GUI to feed the constraint model
- [x] feed the refinement class using the constraint model
- [x] Make sure the popup is drop inside the screen
- [x] Allow to reset to the default values
- [x] Allow to feed the constraint model with the default values (from the refinement class)
- [x] Make sure an already set constraint model is not overwritten by the default values from the refinement class
- [x] Display constrains using the right units
- [x] Rework popup looks and feel (distinguish default and custom constraints)
- [x] Rework main view (distinguish free, fixed and range constraints)

Closes #875


vallsv:
pyfai-range-constraints

kif:
There is probably a mistake sw as I did not manage to calibrate any of the classical images

vallsv:
The fixed constraints was not anymore appyed. Then no fixed wavelength. It can explain your problem.

PR #991, opened 2018-11-05
Update timeout class as explicit

vallsv:
- Fix python2/3 specificity


vallsv:
Should also be backported to a 0.16.1

kif:
Is this related to "inf" in fitting the geometry under python2 ?

vallsv:
Yes it is

PR #995, opened 2018-11-07
992 variance 2d

kif:


kif:
This implements all histogramming based 2D rebinning engines. The LUT-based wll go in another pull-request, this one being already large enough

vallsv:
Looks good

PR #1000, opened 2018-11-15
999 detector pixel corner distortion

kif:
provide a setter which guaranties all detectoe metadata are set properly.


kif:
I'll come back with a proper PR...

Issue #999, opened 2018-11-15
[detector] setting pixel corner is error prone


kif:
Two "failure" of saving the detector position have been reported recently (XFEL + ID15).

PR #1001, opened 2018-11-15
999 detector pixel corner distortion v2

kif:
Provide a setter which properly sets the detector for saving ....


kif:
close #999

vallsv:
Is that normal the method is not called?

vallsv:
LGTM

PR #1004, opened 2018-11-19
fix issue with non-contiguous detector on cython

kif:
geometry.save fixes also the datector name which is now "nexus" instead
of the alias


vallsv:
LGTM

PR #1008, opened 2018-11-20
Fix python2

vallsv:


kif:
running tests on the BL computer

PR #1009, opened 2018-11-21
Constraints plot views

vallsv:
pyFAI contains many plots

This PR sync, the view, the y-direction, the interaction mode, the keep aspect ratio of all of them.


vallsv:
I will create a more general PR to avoid the spam,
spam,
spam,
spam

PR #1011, opened 2018-11-23
Fixed the Cirpad performance issue

alexmarie78:
Cached the computation of pixel corners in order to avoid multiple
computation.


kif:
The second check is useful in case of multi-threaded execution and the calculation was already ongoing.

PR #1012, opened 2018-11-23
Fix fitting regression

vallsv:


kif:
I will test on AGIPD detector ...

kif:
recalib from geometry works, great,
the constrain remains and are taken into account.

but the manual modification of some parameters of the fit, while working OK on the screen turned out to divergent fit as previously. There is still smth fishy around

PR #1019, opened 2018-11-28
Rework the way "no detector" is hilighted

### LINKED ISSUE #1016: [CALIB2] highlight in red "No detector"
on first tab,  third frame.
like "No image"

Nota: if an image is provided, there is a second "No detector" in red but not in the proper section.

Maybe highlight "click here" on the detector side ?

vallsv:
Improve the way no detectors is hilighted.

It could be done in a better way, but that's the simple way to avoid misunderstanding.

Closes #1016


kif:
Looks OK to me.

PR #1022, opened 2018-11-28
Clean up darkflat files

vallsv:
Remove `set_darkfiles` and `set_flatfiles` from `Detector` and from  `AzimuthalIntegrator` classes.

Mostly because it makes my life really difficult to clean up integration configuration.

This methods contains file input, and hardcoded logic to average files. It should not be the responsibility of low level object like that. Fortunately nobody was using metadata relative to it, `Worker` was storing it on it's own. `Detector` still contain a single `flat` and `dark`.

This could break application using pyFAI. But it is much cleaner.


vallsv:
Close it as it will be part of a bigger refactoring

PR #1025, opened 2018-11-30
1021 preproc inlined

kif:
First give it a try before cleaning up ...


kif:
The bug was not in the inlining but rather in delta_dummy value not defined by default and check_dummy left to True by default.

kif:
The inlined version is better from some people knowing the C better than me... so I leave it so.

vallsv:
You could remove ` from libc.stdio cimport printf ` if everything is fixed.

vallsv:
I clean up it, was not a big deal.
Nice catch BTW.

PR #1031, opened 2018-12-04
1029 integation method

kif:


kif:
heavy work in progress

kif:
I have fixed all tests ... does it look reasonable to you ?
Off course all the work is done in the azimuthalIntegrator module

vallsv:
There is still some problem with `Method`.
- The dimension is not needed (as it is used why 2 different methods, `integrate1d` and `integrate2d`), else you want to merge that too.
- `Method` can't be used as key to your registry, as `dim` (needed for your key not for `Method`)  and `target` (needed for `Method` not for your key) to not match what you need.
- There is 2 different construction of `Method` on the code, some do not match the name tuple constructor.
- `target` is still part of `IntegrationMethod` but it is not needed
- `.method` is created in the `IntegrationMethod` constructor, but it is not needed
- `cpu`, `gpu`, `any` are not supported, it is needed

Well, I still don't think it is a good idea to register the same methods for all the available OpenCL targets. I don't see the point. It's always the same thing, then it's useless.

But if it's fine on your side, we can merge it as it is. I can clean up other points on my side, cause i have to do things anyway.

kif:
to answer a few remarks:
* One can now filter on cpu, gpu, ...
* I intend to use IntegrationMethod.method as key in the engine-cache of azimuthal intgerator
* the unconsistance of Method  constructor has been fixed

vallsv:
I see for your cache, but you also intend to use `Method` as a user API to your azimuthal integrator. In this order it contains useless fields.

vallsv:
Anyway, LGTM

PR #1032, opened 2018-12-04
Fix guess poni

vallsv:
The first guess from the peak was not using `guess_poni`.

Now it was fixed as it is done in the peak extraction.

Returning the better of 2 passes with and without `guess_poni`.


kif:
It is probably a wise move ... but it does not explain why it failed so miserably on Marco's image

PR #1034, opened 2018-12-08
Prevent size change during iteration

ronpandolfi:
A simple fix for attempting to deepcopy while simultaneously integrating in multi-threaded use case.


kif:
LGTM ...
a shallow copy of a dict should have 0-performance impact

PR #1037, opened 2018-12-11
Warn if the discontinuity is in the azimuthal range

### LINKED ISSUE #839: [calc] Warning when \chi discontinuity is inside the azimuthal range
This produces a result which is different from what the user expects. 
A WARNING is needed at least. 

Discussing about the need of a auto setChiDiscAtZero/Pi is open.

Reported by Saito Kotaro on the ML.

kif:
close #839


vallsv:
LGTM

PR #1048, opened 2018-12-14
Deprecated comma list file

vallsv:
This PR
- deprecate list of filenames provided to detector and azimuthal integrator (it could stay as an easy way to set but not as getter)
- enforce list of string for dark/flat config.

The idea is to use this (json version 2)
```
flat_field: ["foo1.edf", "foo2.edf", "foo3.edf"]
```
instead of this (json version 1)
```
flat_field: "foo1.edf,foo2.edf,foo3.edf"
```

Not tested


kif:
Looks cleaner than previously

PR #1052, opened 2018-12-14
Fix semaphore api

vallsv:
This PR mostly use the standard API to check if there is available semaphore (but also contains convenient fixes).

But i don't take the logic behind, or it was already broken, cause it was already possible to open many time the "run" screen.


kif:
indeed much better

PR #1058, opened 2018-12-14
Ring displacment

### LINKED ISSUE #1055: [calib2] Take a look at wrong ring alignment
Ring location in the fit screen do not exactly match the acquisition image (feed back from ID15A).

If the problem is obvious, it could be nice to fix it.

vallsv:
Here is the current state. It's better, there is maybe still a bug somewhere
screenshot from 2018-12-14 19-53-05

Closes #1055


kif:
If you wish, one can generate super-fine rings by using fake_calibration_image of the calibrant ...
use a very small "W" and leave U and V to zero. Because it is interpolation and not integration, a W too small will result in no signal at all.

kif:
indeed 0.5 pixel is much !

PR #1061, opened 2018-12-17
Avoid move peaklist scrollbar

### LINKED ISSUE #1015: [calib2] Avoid the peak list scrollbar to move when changing ring number

vallsv:
Closes #1015


kif:
This works as expected.

PR #1062, opened 2018-12-17
Fix edf integration

### LINKED ISSUE #1053: [gui-integration] Not possible to integrate edf files
Feedback from Jerome

vallsv:
Closes #1053


kif:
It works as expected

PR #1085, opened 2019-01-03
Rework file config io

vallsv:
This isolate io for file config into the `io` package. It remove duplicated code and provide a safer compatibility with older config version (while dict are mess).

- Isolate `ponifile`read/write (we could do more: ideally `get/set_config` from azimuthalInteg should not contain `version`, as version should only be part of the storage -> the file)
- helper for integration-config (the json file).

### DIFF
--- a/pyFAI/worker.py
+++ b/pyFAI/worker.py
@@ -85,7 +85,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "17/12/2018"
+__date__ = "04/01/2019"
 __status__ = "development"
 
 import threading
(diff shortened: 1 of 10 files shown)


pyFAI/worker.py
```diff
@@ -413,6 +380,8 @@ def process(self, data, variance=None, normalization_factor=1.0, writer=None, me
         if self.azimuth_range is not None:
             kwarg["azimuth_range"] = self.azimuth_range
 
+        print(kwarg)
```
kif: Trailing print !

pyFAI/worker.py
```diff
@@ -519,6 +489,7 @@ def set_config(self, config, consume_keys=False):
             filenames = _read_filenames(filename)
             method = "mean"
             data = _reduce_images(filenames, method=method)
+            print(data)
```
kif: trailing print

pyFAI/worker.py
```diff
@@ -529,6 +500,7 @@ def set_config(self, config, consume_keys=False):
...
```
kif: trailing print

vallsv:
Fixed

PR #1086, opened 2019-01-04
Custom integration method in the gui

vallsv:
- Support all the method features for integration gui and worker
- Bump integration-config to version 3
- Create a dialog to config the methods (custom imp/algo/split)
- Some changes in `method_registry` to make it work

PR #1085 should be merged first.


vallsv:
It can be reviewed, as i only want to rework a label widget.

vallsv:
Done

kif:
I tested quickly the app and apparently it does the job.

I would have a couple of comments:
* in the main window, the order is split/impl/algo while in the select window the order is impl/algo/split (this is inconstant). I thought a lot on this and believe the most important for the user's point of view is "split" as it impacts the the numerical result, followed by "algo" as it affect the amount of memory used. Finally the implementation is the *least* important from a users' point of view.
* I wonder if the *(only 2D)* is a default error message as some combination are not available, regardless if it is 1d or 2d, for example fullsplit+histogram cannot be (usable) in python regardless if it is 1d or 2d. The error message "not available for 2D integration" neither looks constant as it is always displayed for impossible combinations.

vallsv:
- I think it is better to order it from the more general to the more specific. Mostly split is inside the algo which is inside the impl. If you specify it in this order you should have less problem. But there is no problem to order it in another way.
- For the "only 2d" i notice it too. It would be good to check it together, cause i dont see where is the problem.

vallsv:
THe problem with "only 2D" is curiously fixed by 165ff8907a08268a8d4627c6f324ee2f6ee4a299. The fix convert a generator to a tuple. I guess the "*" do not behave the same for both objects. Very dangerous  behaviour. It would be good to avoid the stars then...

vallsv:
Anyway, i patch the code to fix your comments

vallsv:
Also, it would be good to review the tooltips. As there there is more documentation there.

Issue #954, opened 2018-10-19
Upgrade all cython source to Python3 strings


kif:

kif:
this will fix a warning occurring with Cython

kif:
as #1088 has been merged

PR #1089, opened 2019-01-09
1088 numpy 1.16

### LINKED ISSUE #1088: numpy 1.16 and compiler options ...
We provide a compiler option for making the C-namespace safer. This option is no more compatible  with numpy 1.16
Thanks Fred for spotting the bug.

kif:
close #1088


vallsv:
Should also be reviewed by @t20100 as it will be propagated to silx

kif:
close #954 
close #1088

t20100:
Issue with `numpy` looks to be related to a change of the call to the compiler (see https://github.com/numpy/numpy/commit/83b03407053aefd9553cf3636860f2be8f57f596).

I have a solution that keeps the compile flags and looks to work (still need to be tested on manylinux..).
Those flags are there for manylinux build to prevent symbols to be overridden from one compiled module to another.

t20100:
Commit 650f4b6 makes compile flags compatible with numpy 1.16.0rc1|2.

vallsv:
We should maybe split the numpy problem into another PR?
@t20100 do you think it is better to wait before merging it? As they could change again things?

t20100:
As you want. I would merge and eventually rollback if they change their mind.
To me new way is better.

t20100:
I added a fix for Windows when using OpenMP (from PR https://github.com/silx-kit/silx/pull/2326).

PR https://github.com/silx-kit/silx/pull/2326 also make use of `--pre` for build if you are interested.

kif:
On Fri, 11 Jan 2019 00:09:05 -0800
Thomas VINCENT <notifications@github.com> wrote:

> As you want. I would merge and eventually rollback if they change their mind.
> To me new way is better.

Issue over ... 1.16 is out now 


-- 
Jérôme Kieffer
tel +33 476 882 445

vallsv:
Let's merge it then

kif:
On Tue, 15 Jan 2019 00:42:42 -0800

> Merged #1089 into master.

There are still seg-faults on python version 2.7 andd 3.5 (i.e. numpy 1.16). This is why I suspect this aligment issue.
I tested windows py37 and different flavours of linux ... the aligment
of lut_t (i.e. the struct) and lut_d (the numpy object) are the same.

-- 
Jérôme Kieffer
tel +33 476 882 445

vallsv:
The master is anyway broken if numpy 1.16 is out.

Issue #1091, opened 2019-01-10
Warning in tests


kif:
arnings:/usr/lib/python3/dist-packages/pyFAI/test/test_goniometer.py:106: DeprecationWarning: Please use assertTrue instead.

PR #1093, opened 2019-01-14
[warning] in tests

alexmarie78:
Tried to fix : Warning in tests #1091

### DIFF
--- a/pyFAI/test/test_goniometer.py
+++ b/pyFAI/test/test_goniometer.py
@@ -103,7 +103,7 @@ class TestTranslation(unittest.TestCase):
         g = Goniometer([1., 2., 3., 4., 5., 6., 7.], self.gt, "pilatus100k")
         fname = os.path.join(UtilsTest.tempdir, "gonio.json")
         g.save(fname)
-        self.assert_(os.path.exists(fname), "json file written")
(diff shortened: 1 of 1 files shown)


pyFAI/test/test_goniometer.py
```diff
@@ -103,7 +103,7 @@ def test_goniometer(self):
         g = Goniometer([1., 2., 3., 4., 5., 6., 7.], self.gt, "pilatus100k")
         fname = os.path.join(UtilsTest.tempdir, "gonio.json")
         g.save(fname)
-        self.assert_(os.path.exists(fname), "json file written")
+        self.assertTrue(os.path.exists(fname))
```
kif: why did you remove the comment on the error ?

alexmarie78: I looked upon assertTrue documentation, and I thought that if it is true, the comment on the error would not be printed. My mistake sorry..

vallsv: Usually comment are useless. That's time consuming for nothing. You anyway have to go back to the code.

For example here, the message should be "json file not written"

vallsv:
LGTM

Issue #1094, opened 2019-01-14
Numpy 1.16 and seg-faults 


kif:
Tocheck: memory alignment of complex types (like lut dtype) have changed, investigate this !

kif:
To be integrate in PR#1089

kif:
Funny thing: tests passes in Python3.4 using numpy 1.16 while numpy 1.16 dropped python 3.4 !

kif:
Thanks for the precision. 
We best continue on 
https://github.com/numpy/numpy/issues/12750

kif:
This could be a bug on the Travis side as the 14.04 images (Trusty, gcc 4.8)  are crashing but not the 16.04 images (Xenial, gcc5.3). I suspect hence a compiler issue. We are patching especially against bugs in gcc4.1 used in manylinux ...

vallsv:
Closed by #1096

PR #1096, opened 2019-01-15
raise if alignment is wrong !

kif:
related to #1094


kif:
TODO:
build a xenial VM to build the pyopencl wheel for py37.

PR #1104, opened 2019-01-18
Fixed the 3D preview of CirPad detector.

alexmarie78:
The loading was freezing, and we get this error : 
`ERROR:pyFAI.gui.dialog.Detector3dDialog:incompatible shape for a non-contiguous array`
Now it works with this change.


vallsv:
LGTM

vallsv:
I will rework it in a following patch.

PR #1105, opened 2019-01-18
Improve 3d preview

vallsv:
- Create another 3D toolbutton to display the sample stage (on the fitting task)
- Better error message in case of fail
- Force contiguous array on the CirPAD


vallsv:
I will merge it with my polishing PR

PR #1106, opened 2019-01-21
Definition of CirPad.

alexmarie78:
Hello,
I've recently tried to fix the CirPad definition.
Thanks to the 3D preview, we could figure out that the definition was wrong.
Thus we fixed it according to what we see in 3D.
We are still improving it, but at the moment this fit best with CirPad.


alexmarie78:
Yes, sorry for the omission.

vallsv:
There's no problem.
Let's merge it.

PR #1112, opened 2019-01-24
Typo preventing build on debian8 (numpy 1.8)

kif:
shouldd fix nightly build on edna-site


vallsv:
Why dont you use an equal? Isn't is a specific to numpy `1.12.something`?

kif:
On Thu, 24 Jan 2019 14:54:57 +0000 (UTC)

> Why dont you use an equal? Isn't is a specific to numpy `1.12.something`?

It fails also for 1.8 and 1.12.x

PR #1113, opened 2019-01-25
Added Cirpad2Module and Cirpad2 

alexmarie78:
Hello.
We made new classes to improve the definition of the Cirpad.
It is not the reality, but its one good approach.
We are still trying to improve it, but with this definition, the fitting geometry seems to works better than before.

### DIFF
--- a/pyFAI/detectors/_imxpad.py
+++ b/pyFAI/detectors/_imxpad.py
@@ -40,8 +40,9 @@ __date__ = "18/01/2019"
 __status__ = "production"
 
 import functools
-import numpy
 import json
+import numpy
+from collections import OrderedDict
 from ._common import Detector
 from pyFAI.utils import mathutil
 from .. import geometry
@@ -767,6 +768,17 @@ class Cirpad2(Detector):
             deltaY -= 0.0017
             # deltaZ -= numpy.sin(numpy.deg2rad(-i*6.74))
             # deltaY -= numpy.cos(numpy.deg2rad(-i*6.74))
+
+    def get_config(self):
+        """Return the configuration with arguments to the constructor
+        :return: dict with param for serialization
+        """
+        return OrderedDict((("distance", self.dist),
+                            ("poni1", self.poni1),
+                            ("poni2", self.poni2),
+                            ("rot1", self.rot1),
+                            ("rot2", self.rot2),
+                            ("rot3",self.rot3)))
 
     def _calc_pixels_size(self, length, module_size, pixel_size):
         size = numpy.ones(length)
--- a/pyFAI/detectors/_imxpad.py
+++ b/pyFAI/detectors/_imxpad.py
@@ -44,6 +44,7 @@ import numpy
 import json
 from ._common import Detector
 from pyFAI.utils import mathutil
+from pyFAI.geometry import Geometry
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


kif:
I think there is actually a circular dependency !

vallsv:
The right syntax is `from .. import geometry`

vallsv:
You still have a problem with the tests. Your detector have to be instantiated using `get_config` result.

```
detector = Cirpad2(...)
detector2 = detector.__class__(**detector.get_config())
```

In order to do that, to can override `get_config` or fix the Cirpad2 constructor.

alexmarie78:
Tried to override the get_config function : 
```    
def get_config(self):
        """Return the configuration with arguments to the constructor
        :return: dict with param for serialization
        """
        return OrderedDict((("pixel1",self._pixel1),
                            ("pixel2", self._pixel2),
                            ("distance", self.dist),
                            ("poni1", self.poni1),
                            ("poni2", self.poni2),
                            ("rot1", self.rot1),
                            ("rot2", self.rot2),
                            ("rot3",self.rot3)))
```
Does it look good for you ?

vallsv:
That's up to you. It looks to be consistent with your constructor.

But do you need anyway to custom any of those parameters? I mean i guess the pixel size will never change, it's a physical constraint. I don't know for other parameters. You can change it later if you like, anyway.

alexmarie78:
I guess I can remove the pixel1 and pixel2 but others parameters should be customable as we put different parameters per ImXPadS70.

kif:
Few comments:
You are doing degree <-> rad conversion all the time.
I have designed pyFAI to use "SI" units eveywhere. Can't you define the constant 6.4 degree and store it in radian immediately ?
This would remove this kind of "boiler plate" code from where the complicated part is into the "definition" of the class.

alexmarie78:
I'm only using it once, in the __init__ of the Cirpad2, I think you are looking at Cirpad's class, not Cirpad2.
Still, I changed it in the init of Cirpad2 as you asked : 

``` 
self.rotation.append(numpy.deg2rad(-i*6.74))
```

I still can't manage to understand why : 

```
from .. import geometry
```

doesn't work when it should..

vallsv:
THat's a problem with Python2, ill send you a patch.

vallsv:
Here is the PR on your branch https://github.com/alexmarie78/pyFAI/pull/1

vallsv:
BTW, 2 comments:
- in PyFAI a detector is already a geometry. Then, instead of storing a geometry on one side and a `Cirpad2Module` on the other side, you could merge both and it will fix this cyclic import, cause you don't need to create a geometry anymore.

Instead of 
```
            module = _Cirpad2Module()
            mdgeometry = geometry.Geometry(dist=dist, poni1=poni1, poni2=poni2,
                                           rot1=rot1, rot2=rot2, rot3=rot3,
                                           pixel1=pixel1, pixel2=pixel2, detector=module)

```
You could do
Instead of 
```
            module = _Cirpad2Module()
            module.dist = dist
            module.poni1 = poni1
            ...
```

- Last comment: Each modules should have a different geometry no? Then the Cirpad2 constructor should have dist*20 + poni1*20... no?

That's up to you anyway. I don't know this detector, and the way you calibrate it.

vallsv:
I did not fix other errors.

And if you dont yet know it, you can test using:
```
python2.7 ./run_tests.py pyFAI.test.test_detector.TestDetector.test_detector_instanciate
```

vallsv:
You was not able to import Geometry because there is a dependency. A detector is a geometry by inheritance.

kif:
On Tue, 29 Jan 2019 05:33:53 -0800

> You was not able to import Geometry because there is a dependency. A detector is a geometry by inheritance.

It is not inheritance but composition ... but the idea is there.

vallsv:
Should we close this PR then? As another one exists now.

alexmarie78:
Hello, sorry for the response time, but yes I think this one can be closed !

PR #1122, opened 2019-01-30
Prepare workshop

kif:
To be merged when all features are present


kif:
I made a terrible mistake in incrementing the wrong digit

PR #1132, opened 2019-02-03
new version of PR#1122 ... incrementing the proper value

kif:
should be better now


vallsv:
That's the same after all no?

PR #1133, opened 2019-02-04
Polishing

vallsv:
- Remove pointless warning while integrating


kif:
Looks good ... let's wait for the tests.

PR #1134, opened 2019-02-04
Polishing

vallsv:
- Fix multiframe to HDF5

Previously everything was written at the frame 0


kif:
LGTM ... waiting for tests

PR #1158, opened 2019-02-10
close #1157

### LINKED ISSUE #1157: pip install on non supported architecture
Same as https://github.com/silx-kit/fabio/issues/288

kif:
close #1157

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@
 # ###########################################################################*/
 
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "05/02/2019"
+__date__ = "12/02/2019"
 __status__ = "stable"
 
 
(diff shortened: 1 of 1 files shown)


vallsv:
As i said on fabio, i dont think it's a good place to do it.

And i did not notice a problem with pip, but i notice a -problem on fabio.
Then maybe it is not needed for pyFAI (is it?), And maybe it could be fixed on another way on fabio.

kif:
On Mon, 11 Feb 2019 16:43:33 +0000 (UTC)

> As i said on fabio, i dont think it's a good place to do it.

Nothing enforces setup.py to be called from inside the source
directory ... This is what occures on unsupported platforms.

I agree with you on reverting the patch of sys.path after importing the
module.

 
> And i did not notice a problem with pip, but i notice a -problem on
> fabio. Then maybe it is not needed for pyFAI (is it?), And maybe it
> could be fixed on another way on fabio.

The source code being the same, it will have the same issue for all
three projects. Fabio being installed first, it fails first :)

vallsv:
Thanks for the fix

PR #1161, opened 2019-02-12
1153 drop testimages

kif:
This addresses (partially) the bug encountered in docker container writing outside of the container (/tmp)


vallsv:
LGTM

PR #1171, opened 2019-02-25
Fix tests with --no-opencl

vallsv:
`run_tests --no-opencl` was not anymore working

Maybe since silx opencl integration?


kif:
Looks good.

PR #1184, opened 2019-03-15
Synchronize setup.py with that of silx

t20100:
This PR synchronizes the setup.py with the current one from silx.
In particular:

- `sdist` now build the cythonized files
- The `numpy` version is was compiled with is advertise as a minimum required version
- `cython` is no more in setup_requires (we could argue about that...)


vallsv:
Thanks a lot

PR #1185, opened 2019-03-18
Fix sdist

t20100:
This PR fixes the `sdist` target of setup.py by defining `HAVE_OPENMP` to False while building the tarball.


vallsv:
LGTM

PR #1186, opened 2019-03-18
An alternative approach to OpenMP support in histogram

t20100:
This PR proposes to move the choice of which implementation of the histogram to use from cythonization phase to C compilation phase.
To do so, both implementations of the histogram are built as separated Cython extensions (using stubs for OpenMP if it is not available).
The choice of implementation in the `histogram` module is done at runtime depending whether it was built with or without OpenMP.
There is a breach there as on can build the `_histogram*` modules without OpenMP and then the `histogram` module with OpenMP.... This can be (and should be) prevented, by checking if the `_histogram_omp` module was built with OpenMP. The `histogram` module can then be a pure Python module. If you are interested I can update this PR for it.

This looks to work, but requires more testing... Don't hesitate to close if not appropriate !

### DIFF
--- a/pyFAI/ext/_histogram_nomp.pyx
+++ b/pyFAI/ext/_histogram_nomp.pyx
@@ -34,9 +34,11 @@ __license__ = "MIY"
 __copyright__ = "2011-2016, ESRF"
 __contact__ = "jerome.kieffer@esrf.fr"
 
+include "regrid_common.pxi"
+
 from libc.math cimport floor, sqrt
 import logging
-logger = logging.getLogger(__name__ + "_nomp")
+logger = logging.getLogger(__name__)
 
 
 @cython.cdivision(True)
--- a/pyFAI/ext/_histogram_omp.pyx
+++ b/pyFAI/ext/_histogram_omp.pyx
@@ -45,6 +45,8 @@ import logging
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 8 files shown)


t20100:
While at it, improved the check of histogram built with OpenMP at runtime.
It now relies on the compilation of `_histogram_omp`... `histogram` is now a pure Python module.

vallsv:
Thanks.
That's much better than what i expect.

vallsv:
I tested that:
- `pyFAI.ext.histogram` load the right module on both cases (`--no-openmp` `--openmp`)
- And timeit `pyFAI.ext._histogram_omp` with and without openmp (histograming from 1.30ms to 500µs)

Then it looks to work fine. Thanks a lot, i think it is a very nice feature. We could add strategies according to data size or things like that. Or disable OpenMP to use multiprocess... This kind of stupid tweaks.

BTW `_histogram_nomp` looks always slower than `_histogram_omp` with or without OpenMP. But i guess my test was not representative (the one from `test_histogram`).

kif:
On Mon, 18 Mar 2019 16:59:59 +0000 (UTC)

> BTW `_histogram_nomp` looks always slower than `_histogram_omp` with or without OpenMP. But i guess my test was not representative (the one from `test_histogram`).

This sounds strange to me as omp adds quite a subtential overhead (not
only in numbre of lines of code).

PR #1187, opened 2019-03-18
Add *.cpp to manifest

t20100:
This PR adds `*.cpp` to `MANIFEST.in` as it was missing while there is some C++ extensions in `pyFAI.ext`.


vallsv:
LGTM

PR #1198, opened 2019-04-16
1189 doc geometry

kif:
This is basically the document provided by Carsten to explain how he built the translator from ImageD11 to pyFAI and vice-versa. The document has been translated by pandoc from latex and typos have been manually fixed in the RST.


vallsv:
Looks ok.

PR #1203, opened 2019-04-23
Provide an history of geometry

vallsv:
Provide an history of previous geometry.

This geometries are stored everytime after manual actions (fit, reset, param edition).

Selecting a previous entry from the history set the current display to this new value.

Todo:

- [ ] Maybe put a size limit to the history? (remove older, or bad RMS) (not a big issue, as it is not supposed to stay open many days)
- [x] Rework the computation of the RMS (now the stored one looks buggy time to time)
- [x] Rework the processing based on update of `CalibrationState`
- [ ] Feed `fittedGeometry` from the update of `CalibrationState`
- [ ] Rework the fitting/ring computation/rms computation using a thread
- [ ] Allow to cancel computation if something else is edited (manual edition should be non-blocking)

Screenshot from 2019-04-23 11-24-49


vallsv:
This PR is not yet ready as some improvements still have to be done, but it is testable.

vallsv:
It can be merge as it is.
I would like to do more, but maybe i will not have anymore time.

kif:
From what I have tested ... it works.

PR #1207, opened 2019-04-25
Histogram engine rebased

kif:
This PR provides a second integration engine based on numpy histograms. It validates the error propagation chi^2 test


vallsv:
BTW, is the variance notebook have to be provided with it's output?

Issue #1208, opened 2019-04-30
Geometry: Pixel coordinate from tth and chi


tifuchs:
Hello, 
is there a function to calculate the pixel coordinate from given tth and chi (or Q, r, ...)? 
I.e. analytically get the inverse of Geometry.chi and Geometry.tth?

I would like to have something like this for single crystal scattering at ID31, where I want to calculate the theoretical position of a Bragg-reflection on the detector.

Thank you so much for your great work!

kif:
Hi,

There is something that works a bit like that but uses linear interpolation for the last pixel.
The code is available in pyFAI.ext.invert_geometry:
https://github.com/silx-kit/pyFAI/blob/master/pyFAI/ext/invert_geometry.pyx

Unfortunately, I did not include it into the generated documentation (hence the label "doc").
Basically you initialize it with the radial and azimuthal array one gets from the geometry and then call the instance with the values you are expecting. It can produce a pixel coordinate (integer) or the interpolated assuming a second order polynomial extension of the cost function.

TODO: Update the doc.

tifuchs:
Thank you very much!

I've tested the code and it actually seems to work with any two 2d-arrays. So you can also put in tth and chi as I was asking for.

The only issue is that the code is extremely slow. 
With 1679 x 1475 Detector size my machine needs about 240 ms per position.
Maybe this is something to improve in the future...
(I don't need the speed at the moment)

kif:
On Tue, 30 Apr 2019 15:49:19 +0000 (UTC)
tifuchs <notifications@github.com> wrote:

Indeed it has not been optimized for speed. 240ms looks a lot to me I
guess a factor 10 faster is acheivable.

Maybe I forgot to declare a variable somewhere. In the GUI we don't
call this code very often so it has not benchmarked.

kif:
The performances looks as expected to me :
```
import pyFAI
det = pyFAI.detector_factory("Pilatus2M")
ai = pyFAI.AzimuthalIntegrator(detector=det)
r=ai.rArray()
c=ai.chiArray()
from pyFAI.ext.invert_geometry import InvertGeometry
ig = InvertGeometry(r,c)
%timeit ig(0.2, 0.8)
```
3.42 ms ± 139 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

tifuchs:
I also just noticed that the performance is good on rnice and another Linux machine, that I've tested. 
[...]

PR #1210, opened 2019-05-05
Minimal documentation on invert_geometry

kif:
Addresses the documentation part of #1208.

The performance issue on windows remains a mystery.


vallsv:
Noice

PR #1212, opened 2019-05-06
Simplify the kernel for being able to run on PortableCL with PTX backend

### LINKED ISSUE #1211: [OpenCL]  Simplify the kernel to make azimuthal integration compatible with POCL + PTX
`pown` instruction from OpenCL cannot be directly translated into PTX (Nvidia) instruction. 
As only power 1 and 2 are possible options, using a switch would make the code compatible with POCL on platforms where nvidia does not provide opencl drivers.

kif:
close #1211


kif:
After a quick look at the assembly produced, this patch is likely to speed up substantially the execution

vallsv:
LGTM

PR #1213, opened 2019-05-06
Expose cython histograms

kif:
This is the last implementation of the histogram without pixel-splitting. After NumPy and OpenCL, here is the Cython version.


kif:
Looks better now ?

vallsv:
LGTM

PR #1215, opened 2019-05-07
A couple of small error fixed:

kif:
* The benchmark tool was ending in error when no display was available
* The OpenCL code was still using the *pown* function, not available on certain platforms
* The benchmarking tool has been simplified, removing the LUT implementation as it is not of prime importance
* There has been a bug in the OpenCL device selection since the introduction of the registry. Now the benchmarking tool prints out the actual used device, not only the one it wish to use.


vallsv:
LGTM

PR #1221, opened 2019-05-09
This is workaround for issue #1218.

### LINKED ISSUE #1218: Issue when using rebinning engines outside their "natural" range
Reported by NSLS-II & APS:
Prior of using pyFAI, we are calculating q values. After that we using
the splitBBox.histoBBox2d() with the q range as input.
 
Last week, we run the reduction in a different energy range, leading to
different value ranges for the arguments pos1 and delta_pos1.
And the data reduction start failing when the values for the vertical range
are reaching 3.14 (probably pi), as shown in the pdf file.
 
To go around this problem, I just divide by 10 the range of the input
in splitBBox.histoBBox2d()
and multiply by 10 afterwards and this is working fine.

Is it something you have already observed? And if so, do you have any way
to go around?

kif:
This option should probably be extended to all other rebinning engines but this a lot more of work.
@ronpandolfi, @gfreychet can you confirm it works for your use-case ?
close #1218

### DIFF
--- a/pyFAI/ext/splitBBox.pyx
+++ b/pyFAI/ext/splitBBox.pyx
@@ -32,7 +32,7 @@ Splitting is done on the pixel's bounding box similar to fit2D
 
 __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.kieffer@esrf.fr"
-__date__ = "02/05/2019"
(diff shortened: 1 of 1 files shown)


vallsv:
LGTM

PR #1227, opened 2019-05-15
Improve ellipse fit

vallsv:


kif:
Should be much more robust !

PR #1230, opened 2019-05-15
Fix pyFAI-calib2 for debian8

t20100:
This PR makes `pyFAI-calib2` works with `PyQt5` from debian 8 (i.e., v5.3.2).
Feel free to reject if you don't want to support debian 8 (as Debian does).


vallsv:
LGTM

PR #1240, opened 2019-05-16
Add ellipse to binder

vallsv:


kif:
Thomas has an alternative solution which provides this page dynamically. It will be even better.

PR #1244, opened 2019-05-16
Update binder environment

t20100:
This PR generates the binder landing notebook when creating the environment, thus the list of notebook is updated.

To test the PR: https://mybinder.org/v2/gh/t20100/pyFAI/binder?filepath=binder%2Findex.ipynb


vallsv:
Not sure it is a good idea. As if it is supposed to be pedagogic it would be better to manage it manually, grouping them, or providing description...

The index.ipynb could be part of pyFAI instead of part of binder.

t20100:
I defintely agree, a manually maintained `index.ipynb` is definitely better... but it needs to be maintained.
It depends how you want to handle this.

t20100:
Just in case you are interested, I recently added alternative binder service badges (Using https://binder-panosc.fedcloud-tf.fedcloud.eu and https://notebooks.gesis.org) to https://github.com/silx-kit/silx-training

See https://github.com/silx-kit/silx-training/commit/8a8cf844f85fce1e581493a907c746a9a583e184

kif:
Thanks,

I am interested for the upcomming tutorials on pyFAI

PR #1249, opened 2019-05-16
Image d11

kif:
import and export to ImageD11 format


jonwright:
Thanks ! It should be very useful. I will reflect on the Oij flips. For completeness, there is a document to cite about the fable geometry here:

https://sourceforge.net/p/fable/wiki/_discuss/thread/3474b8d3/c8a8/attachment/Fable_Geometry_version_1.0.8.pdf

Wavelength units should match unit cell if you decide to provide one (e.g. from a calibrant used). Typically Angstrom. 

At some point I will try to make some test cases verifying the 3D x,y,z lab co-ordinates from both packages, but not any time soon (ImageD11.transform.compute_xyz_lab).

PR #1263, opened 2019-05-20
Do not compute all the ring polygons but only expected ones

vallsv:
Previous refactoring allow to use marching squares in a better way.

This PR move the marching squares from the fitting processing to the GUI.

As result we can only compute rings which are requested by the GUI, instead of computing everything.

This speed up each steps of the geometry fitting per 2 when using a very big image.


vallsv:
Error is about the logs. It is safe to merge it.

kif:
LGTM

PR #1270, opened 2019-06-07
1269 ocl integrator

### LINKED ISSUE #1269: New generation integrator: OpenCL & sparse matrix multiplications
Among the 27 methods to re-implement this is probably the most useful one.
Implement it twice:
* signal/normalization arriving as float2 -> for XPCS data reduction. Maybe as a function to be used twice later on
* signal/variance/normalization/count arriving as float4 for general purpose integration. Goes on step further with average intensity calculation and standard deviation calculation

kif:
This PR provides integrator with proper variance propagation
close #1269 
The work goes on in 5 subsequent steps phases:
* [X] Code the OpenCL kernel
* [x] Interface in Python the OpenCL code to make it an "engine"
* [x] Provide test to this engine
* [x] Make the engine available via the AzimuthalIntegrator and selectable via method
* [x] Test this new method


kif:
This is again a big piece of work ... Let's check if it does not create too much regressions

PR #1279, opened 2019-08-27
Added an ImXPadS70 in vertical shape

alexmarie78:
Hello there,

To deal with the work on the cirpad detector, we need a new class wich is a vertical ImXPadS70 in pyFAI.
In fact it represents well one of our 20 modules of the Cirpad detector thus, it could helps us a lot if you actually merge it into pyFAI for the next release.

Thanks in advance,

Alexandre Marie


kif:
LGTM

Issue #1281, opened 2019-09-20
[pyFAI] Error observed when using pyFAI-calib2


picca:
Hello, I am using the current dev of pyFAI and I get this kind of errors.

```
picca@diffabs6:~/notebooks-soleil/cirpad$ emacs Soleil_Diffabs_Calibration_Cirpad.org
Final residual: 1.891020571821856e-06 (after 1 iterations)
Final residual: 2.0654419112680463e-09 (after 1 iterations)
Final residual: 3.7160486130402986e-09 (after 1 iterations)
ERROR:pyFAI-calib2.UNCAUGHT:float() argument must be a string or a number, not 'NoneType'
Traceback (most recent call last):
  File "/home/experiences/instrumentation/picca/pyFAI/build/lib.linux-x86_64-3.7/pyFAI/gui/tasks/GeometryTask.py", line 969, in __geometryCustomed
    calibration.fromGeometryModel(geometry, resetResidual=True)
  File "/home/experiences/instrumentation/picca/pyFAI/build/lib.linux-x86_64-3.7/pyFAI/gui/helper/RingCalibration.py", line 425, in fromGeometryModel
    self.__geoRef.poni2 = model.poni2().value()
  File "/home/experiences/instrumentation/picca/pyFAI/build/lib.linux-x86_64-3.7/pyFAI/gui/helper/RingCalibration.py", line 74, in __setattr__
    return object.__setattr__(self.__geoRef, name, value)
  File "/home/experiences/instrumentation/picca/pyFAI/build/lib.linux-x86_64-3.7/pyFAI/geometry.py", line 2092, in set_poni2
    self._poni2 = float(value)
TypeError: float() argument must be a string or a number, not 'NoneType'
Final residual: 4.337900633767923e-09 (after 1 iterations)
ERROR:pyFAI-calib2.UNCAUGHT:'<' not supported between instances of 'GeometryRefinementContext' and 'GeometryRefinementContext'
Traceback (most recent call last):
  File "/home/experiences/instrumentation/picca/pyFAI/build/lib.linux-x86_64-3.7/pyFAI/gui/tasks/GeometryTask.py", line 864, in __initGeometry
    self.__initGeometryFromPeaks()
  File "/home/experiences/instrumentation/picca/pyFAI/build/lib.linux-x86_64-3.7/pyFAI/gui/tasks/GeometryTask.py", line 830, in __initGeometryFromPeaks
[...]

PR #1282, opened 2019-09-20
[calib2] Only use the first element to sort fits

vallsv:
I did not test it.

I only test the piece of code using ipython (Python3)

Relative to #1281


picca:
I just checked, and it solve my problem :)) thanks a lot

kif:
Thanks

PR #1284, opened 2019-10-02
splits instead of slits typo

kif:
tests ongoing


vallsv:
Wider that expected.... LGTM

PR #1293, opened 2019-11-19
1292 separate container rebased

kif:
Apparently eclipse@home is more picky about trailing white spaces and CR.

@nils-blanc could you check that you can retrieve the median-filtered signal as expected ?


kif:
I flagged this as SSX as the *separate* method is the core of the Bragg-spot extractions for single crystal data.

kif:
@vallsv as you initiated this way of doing, can you confirm it fits your architecture to keep it consistent ?

vallsv:
There is nothing special here. It was not mine, but decided at a silx meaning + retro-compatibility.

This said it looks fine. I think the most important thing is the non-mutability, and it's the case for the public API, then it's fine.

You want me to merge it?

vallsv:
Lets go.

PR #1297, opened 2019-12-06
AttributeError when calling AbstractCalibration.preprocess()

bertrand-faure:
'Massif' object has no attribute 'setValleySize'

It seems that this method has been removed from the Massif class, so this PR sets the value of valley_size directly to avoid an AttributeError.


kif:
Well spotted. Thanks for the contribution

PR #1301, opened 2019-12-13
Added Imin in the extract_cp method.

picca:
Sometimes, the extraction of peaks from an image gives lot's of wrong
control points. We observe this phenomena when we have lot's of big
crystals (non continuous rings). This parameter allows the user to
solve this issue, by tuning the Imax expected for a real peak.


kif:
Looks good,
Don't worry about the tests not passing, there are related to scipy 1.4 which is arriving (the issue is fixed in another PR).

Thanks for your contribution

PR #1305, opened 2020-01-15
1304 hdf5plugin

### LINKED ISSUE #1304: Ensure scripts try to use hdf5plugin to access their data

kif:
close #1304

### DIFF
--- a/pyFAI/app/__init__.py
+++ b/pyFAI/app/__init__.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+#    Project: Fast Azimuthal integration
+#             https://github.com/silx-kit/pyFAI
+#
+#
+#    Copyright (C) 2013-2020 European Synchrotron Radiation Facility, Grenoble, France
+#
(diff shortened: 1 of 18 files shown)


vallsv:
I don't think it is a good idea to create an `__init__` module here.

Initializing the logging or loading library is a tricky process.

You will not be able to log early debug message this way.

The problem is not to process this way, but to not allow to process in a different way for one of the applications.

kif:
On Wed, 15 Jan 2020 04:57:48 -0800

> I don't think it is a good idea to create an `__init__` module here.

The __init__.py already existed, but was empty.
 
> Initializing the logging or loading library is a tricky process.

Indeed but __init__ was anyway loaded.
 
> You will not be able to log early debug message this way.

Python has an option to start in debug mode but I did not test if it was actually working. 
 
> The problem is not to process this way, but to not allow to process in a different way for one of the applications.

All applications had the same logging initialization pattern.

vallsv:
There is a typo. `raise ImportError` have to be removed.

kif:
On Mon, 20 Jan 2020 00:33:34 -0800

> There is a typo. `raise ImportError` have to be removed.

:)

vallsv:
LGTM

PR #1308, opened 2020-01-18
set values in FIT2D to SI to use the GUI conversion mechanism

### LINKED ISSUE #1306: Fit2D geometry is wrongly displayed ...
This bug affects only the GUI (integrate and diff-map), the calculations are properly done but not properly displayed. Often the geometry is declared as invalid while it is not.

kif:
Close #1306 

I don't like the conversions performed here ... but it works (I did not check everything).

### DIFF
--- a/pyFAI/app/diff_map.py
+++ b/pyFAI/app/diff_map.py
@@ -34,14 +34,20 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "10/10/2018"
+__date__ = "17/01/2020"
 __satus__ = "Production"
 
-import logging
 import sys
+import logging
 logging.basicConfig(level=logging.INFO)
 logging.captureWarnings(True)
-logger = logging.getLogger("diff_map")
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


vallsv:
The conversion can be done in the widget by description line 95. But i understand fit2d uses mm and degree? And your fid2d API provides then meters and radian?

If it is that, there is a unit for the internal data provided, and a unit for the displayed data. The internal unit have to be consistent with yout API, and the displayed unit have to be consistent with the Fit2D data.

Except that, nice catch.

kif:
> The conversion can be done in the widget by description line 95.

I tried and failed ... this is why I suspect there is smth I don't master (yet?).

> But i understand fit2d uses mm and degree? 

Yes it does.

> And your fid2d API provides then meters and radian?

Nop, they where already in mm and degrees.

> If it is that, there is a unit for the internal data provided, and a unit for the displayed data. The internal unit have to be consistent with yout API, and the displayed unit have to be consistent with the Fit2D data.

It tried to tell GeometryDialog that the provided values were in mm &
deg but as it kept assuming they were in SI, I implemented conversions
in each directions.

The best would be to avoid any conversion (-code) in the dialog and have all this implemented in the associated model.
 
> Except that, nice catch.
:)

vallsv:
I think i found the problem. Let me know when you have time, we can check together.

vallsv:
LGTM

PR #1322, opened 2020-02-17
load control points from file through a context manager

bertrand-faure:
Hi,
I suggest opening the control point file within a context manager to make sure it is closed properly.
Best regards,
Bertrand


kif:
Hi Bertrand, 
You are perfectly right. Any uncaught exception in the parsing would leave the file opened and on Windows this would lock the file.
Thanks for your contribution, I'll review it soon

Issue #1344, opened 2020-04-03
Error when packaging pyFAI with PyInstaller


jbhopkins:
I create packaged versions of my program (https://bioxtas-raw.readthedocs.io/) using PyInstaller (https://pyinstaller.readthedocs.io/). When packaging pyFAI there's an error related to how it's searching for the resource path (there's no utils directory in the PyInstaller pyFAI package). It's an error with one line in `_get_data_path()` in `__init__.py` in `utils`.

Current code (lines 163-166 in the version I have installed):
```python
    resources = [
        os.environ.get("PYFAI_DATA"),
        data_dir,
        os.path.join(os.path.dirname(__file__), "..")]
```

It's a simple fix, you just need to make it an absolute path:
```python
resources = [
        os.environ.get("PYFAI_DATA"),
        data_dir,
        os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))]
```

I'm happy to submit this as a pull request, if you let me know what branch I should base it off of. Or if you just want to make the simple change that might be easiest.

Thanks.

Jesse

kif:
On Fri, 03 Apr 2020 09:36:44 -0700
jbhopkins <notifications@github.com> wrote:

> I create packaged versions of my program (https://bioxtas-raw.readthedocs.io/) using PyInstaller (https://pyinstaller.readthedocs.io/). When packaging pyFAI there's an error related to how it's searching for the resource path (there's no utils directory in the PyInstaller pyFAI package). It's an error with one line in `_get_data_path()` in `__init__.py` in `utils`.
> 
> Current code (lines 163-166 in the version I have installed):
> ```python
>     resources = [
>         os.environ.get("PYFAI_DATA"),
>         data_dir,
>         os.path.join(os.path.dirname(__file__), "..")]
> ```
> 
> It's a simple fix, you just need to make it an absolute path:
> ```python
> resources = [
>         os.environ.get("PYFAI_DATA"),
>         data_dir,
>         os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))]
> ```
> 
> I'm happy to submit this as a pull request, if you let me know what branch I should base it off of. Or if you just want to make the simple change that might be easiest.

Hi Jesse,

You can make a PR on the master-branch by citing this issue. The CI will be triggered automatically.
I may ask Valentin or Fred Picca to have a look at it as they worked on this part.

Cheers,

Jerome

PR #1345, opened 2020-04-03
Fix path bug when packaging with PyInstaller

jbhopkins:
Fixes issue #1344 by changing normalizing relative path into an absolution path.

### DIFF
--- a/pyFAI/utils/__init__.py
+++ b/pyFAI/utils/__init__.py
@@ -163,7 +163,7 @@ def _get_data_path(filename):
     resources = [
         os.environ.get("PYFAI_DATA"),
         data_dir,
-        os.path.join(os.path.dirname(__file__), "..")]
+        os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))]
     try:
         import xdg.BaseDirectory
         resources += xdg.BaseDirectory.load_data_paths("pyFAI")
--- a/pyFAI/utils/__init__.py
+++ b/pyFAI/utils/__init__.py
@@ -163,6 +163,7 @@ def _get_data_path(filename):
     resources = [
         os.environ.get("PYFAI_DATA"),
         data_dir,
+        os.path.join(os.path.dirname(__file__), ".."),
         os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))]
     try:
         import xdg.BaseDirectory


kif:
@picca, Fred, you introduced this part, could you please have a look at this PR. Thanks.

picca:
Hello, resources is a list of path where we search the data files.
I would like to understand why there is no utils path in the pyinstall package ?

jbhopkins:
I don't understand the packaging mechanism of pyinstaller particularly well, but I'll try to describe what happens as I understand it. It packages compiled python bitcode into a single executable archive with some kind of internal structure. Additionally it packages other resources (compiled binaries, images, etc), into adjacent folders. So for pyFAI most of the code ends up in the compiled binary, and you end up with a separate ext and resources folders. Since the utils folder contains only python code it doesn't get packaged as a separate folder (though the code is packaged in the executable archive). 

Normally this isn't an issue, since there's internal structure to the executable archive that pyinstaller creates so the import statements work as expected (I think, or maybe it bootstraps on the import process, I don't really know). However, when it tries to find the resources directory relative to the utils directory, it's looking outside of the executable archive. In that case, there's no utils directory outside the archive, so it can't find the resources directory. If you instead make the path absolute to the resources directory that seems to work.

If you want to know more about what pyinstaller is doing then that (and I may well have gotten something wrong), you'll have to read the documentation:
https://pyinstaller.readthedocs.io/en/stable/advanced-topics.html#the-bootstrap-process-in-detail

If you want to test out the packaging for yourself, you'll need the hooks file I created for pyFAI, which tells pyinstaller what extra files to grab. That's located here:
https://sourceforge.net/p/bioxtasraw/git/ci/python3/tree/installer/

picca:
thanks

are you using this in order to add the data files ?

https://pyinstaller.readthedocs.io/en/stable/spec-files.html#adding-data-files

jbhopkins:
So I use the spec file to add my program's resources as described in that link. I created a hook file for pyFAI to add the pyFAI resources:
https://pyinstaller.readthedocs.io/en/stable/hooks.html

picca:
Inn that case it seems that the best way  to retreave the data is by using the
pkgutil module

https://pyinstaller.readthedocs.io/en/stable/spec-files.html#adding-data-files

Do you think that we should fallback to this pkgutil is all the previous search failed ?

jbhopkins:
I've never used pkgutil myself, so I'm not sure I can say. It is a standard lib module, so it's not like you'd be adding extra dependencies if you wanted to use it. From the documentation it seems like it might be suitable.

https://docs.python.org/3/library/pkgutil.html

picca:
I just says that this  how to add and use the data from the pyinstall documentation :)

this get_data was added a long time ago in Python

https://www.python.org/dev/peps/pep-0302/

Nevertheless, if you just want to add your quick fixup, instead of replacing the relative path, just add one more whcih is the once with the abs path.
This way it dos not change the behaviour of pyFAI, but just deal with your  case.

I am not sure that the current remativ path is a bug.

cheers

jbhopkins:
I modified the pull request to add an additional path, rather than eliminating the relative path. I could try adding the pkgutils approach if you like, but you'd need to provide me with some instructions on how to install pyFAI for development (I had tested my previous small change just by directly changing the code in the installed package, which is not good practice). I tried to do an editable install with pip (basically jut following the matplotlib contributors guide: https://matplotlib.org/devel/contributing.html), but that failed with an error about running from a built package not source.

kif:
I merge this as no side effect is expected as both relative and absolute path are provided.
Concerning the contribution guide, I'll think on it.

PR #1346, opened 2020-04-06
Complete contributing

kif:
@jbhopkins does this help ?


jbhopkins:
That's essentially what I was looking for. The bootstrap.py is the main thing I was missing. I guess a little more clarification for those of us not used to using the library might be helpful. 1) Do I have to rebuild every time I make a code change (even if it's in pure python code, not compiled code)? 2) Can I run scripts using pyFAI using ./boostrap.py <scriptname>? I'm assuming so, because you say it will run the GUI but clarification there would be great.

jbhopkins:
With the changes, looks good to me. It's certainly enough I could do my own modifications and testing.

PR #1348, opened 2020-04-23
close #1347

### LINKED ISSUE #1347: mask_crc reset unconditionally for "lut" method in _integrate1d_legacy
See https://github.com/silx-kit/pyFAI/blob/master/pyFAI/azimuthalIntegrator.py#L1073

Is there a reason for this? This has put a huge penalty when processing multi frame images.

kif:
Sorry for Eclipse which reformatted the code with.


kif:
This was a very old piece of code (8 year old) and I could not find any justification for it.

PR #1353, opened 2020-05-18
Enforce target to be a tuple

kif:
The rational: json de-serialisation creates lists. here we need tuples
for immutability.


vallsv:
LGTM

PR #1357, opened 2020-06-15
Accept scalar for flat/dark

### LINKED ISSUE #1356: Support dark/flat being constant
In some detectors, it can be useful to set the dark/flat as a constant.
-->ID02

kif:
Previously it was crashing at the CRC calculation as scalar values are not implementing
the buffer interface.
close #1356


vallsv:
LGTM, but this patch relative to numpy is really sad.

PR #1362, opened 2020-06-22
Simplify the calculations if the detector declares itself as distortion-free

### LINKED ISSUE #1361: Distortion correction for detector not needing distortion correction
While this is stupid and wasting resources, I tested to correct data from a detector having no distortion (test case for #1360).
Due to numerical noise (the calculation are performed in 32bit float), there are 30% of the pixel which get the contribution of a neighboring pixel with a weight less than 3e-5 for the pilatus100k.

Suggested patch: Detector without distortion should see their correction matrix to be identity!
```
import pyFAI
from pyFAI.distortion import Distortion
p = pyFAI.detector_factory("Pilatus100k")
%timeit sparse.identity(numpy.prod(p.shape), format="csr") --> 179µs
%timeit csr = Distortion(detector=p).calc_LUT() --> 13ms
```

kif:
close #1361 
This code deserves more testing :(

### DIFF
--- a/pyFAI/distortion.py
+++ b/pyFAI/distortion.py
@@ -4,7 +4,7 @@
 #    Project: Azimuthal integration
 #             https://github.com/silx-kit/pyFAI
 #
-#    Copyright (C) 2013-2018 European Synchrotron Radiation Facility, Grenoble, France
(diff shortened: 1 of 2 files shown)


vallsv:
I can say with a single look that it is bug free

PR #1363, opened 2020-06-23
Specify some version number for supporting python 3.5

kif:
Select elder version of pytools when using python 3.5

As this is for CI, I did not open an issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
       - ocl-icd-libopencl1
       - opencl-headers
       - libnuma1
-      - ocl-icd-dev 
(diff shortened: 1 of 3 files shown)


vallsv:
I still don't understand why numpy version have to be specified. That's a numpy issue?

kif:
On Tue, 23 Jun 2020 04:07:23 -0700

> I still don't understand why numpy version have to be specified. That's a numpy issue?

I think so.

Maybe this has been fixed with the final revision.
Here it is not numpy but pytools. I opened them an issue.

Thanks for the merge

Issue #1366, opened 2020-06-25
[pyFAI-calib2] issue with calibration (probably the display of the ring after the fit)


kif:
Context: calibration of a 3D detector (i.e. not flat) loaded from a Nexus file where binning (2x2)  is also applied which would explain the broadcasting error.
```
ERROR:pyFAI-calib2.UNCAUGHT:operands could not be broadcast together with shapes (3008000,3) (12032000,3) 
Traceback (most recent call last):
  File "/opt/VirtualEnv/py37/lib/python3.7/site-packages/pyFAI/gui/tasks/GeometryTask.py", line 972, in __geometryCustomed
    state.update(calibration)
  File "/opt/VirtualEnv/py37/lib/python3.7/site-packages/pyFAI/gui/tasks/GeometryTask.py", line 220, in update
    self.__poni = calibration.getPoni()
  File "/opt/VirtualEnv/py37/lib/python3.7/site-packages/pyFAI/gui/helper/RingCalibration.py", line 389, in getPoni
    solidAngle = self.__geoRef.solidAngleArray(shape=self.__image.shape)
  File "/opt/VirtualEnv/py37/lib/python3.7/site-packages/pyFAI/geometry.py", line 1150, in solidAngleArray
    shape, dtype=numpy.float32)
  File "/opt/VirtualEnv/py37/lib/python3.7/site-packages/numpy/core/numeric.py", line 1809, in fromfunction
    return function(*args, **kwargs)
  File "/opt/VirtualEnv/py37/lib/python3.7/site-packages/pyFAI/geometry.py", line 1122, in diffSolidAngle
    cosa = self._cached_array["cos_incidence"] = self.cos_incidence(d1, d2)
  File "/opt/VirtualEnv/py37/lib/python3.7/site-packages/pyFAI/geometry.py", line 1078, in cos_incidence
    return (t * orth).sum(axis=-1).reshape(d1.shape)
ValueError: operands could not be broadcast together with shapes (3008000,3) (12032000,3) 
```

Example files and test case by email

kif:
The calibration with the CLI worked like a charm, I didn't use it for a year.

kif:
I managed to reproduce the bug without the GUI

kif:
The detector config does not contain the binning information

vallsv:
So the created detector object do not fit the data? That's just that?
That's something which should be anyway protected in both applications.

kif:
[...]

PR #1367, opened 2020-06-25
Define another cylindrical detector from Rigaku

kif:
Minor refactoring of the detectors distribution in the files


kif:
I will put the solution of #1366 as part of this PR, they are directly related.

PR #1369, opened 2020-06-26
Refactor Sparse integrator written in Cython

kif:
There are 2 types of sparse integrators, LUT and CSR ones. All of them implement a legacy integrate and a new generation integrator. Hence it makes sense to make them all derive from common ones which implement actually the integration.
All existing classes doing this can then derive from those classes.

Distortion correction could also be derived from those classes. 

This PR addresses only 1D intergation in the first instance, as this PR is likely to be pretty large


kif:
As a refactoring, I expect this PR to reduce the number of lines ... for now it is rather the opposite :(

PR #1379, opened 2020-07-15
Remove the reference to dx and dy as they are not used.

### LINKED ISSUE #1378: setting detector distortion array to same size as image causes ValueError
hi,

when I try to set dx and dy for a detector with arrays of the same size as the image, a ValueError is risen in the calc_cartesian_positions function:

Example code which reproduces the error:
```
import numpy as np
from pyFAI.detectors import Detector
from pyFAI.calibrant import CalibrantFactory
factory = CalibrantFactory()

img_shape=(2880, 2880)

detector = Detector(150e-6, 150e-6)
detector.max_shape = detector.shape = img_shape

dx = dy = np.ones(shape=img_shape)

detector.set_dx(dx)
detector.set_dy(dy)

from pyFAI.geometryRefinement import GeometryRefinement

pattern_geometry = GeometryRefinement([[1, 1, 0],[2,1,1]],
        dist=1,
        wavelength=0.3344e-10,
        detector=detector,
        calibrant=factory('CeO2'))

```

kif:
close #1378 

@CPrescher could you have a look if it fits your needs ?

### DIFF
--- a/pyFAI/detectors/_common.py
+++ b/pyFAI/detectors/_common.py
@@ -210,8 +210,6 @@ class Detector(with_metaclass(DetectorMeta, object)):
         self._maskfile = None
         self._splineFile = None
         self.spline = None
-        self._dx = None
-        self._dy = None
         self._flatfield = None
         self._flatfield_crc = None  # not saved as part of HDF5 structure
         self._darkcurrent = None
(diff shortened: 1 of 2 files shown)


CPrescher:
Hi!

thank you for your fast response and change. I think it looks good. However, I think the current implementation has an issue if you call set_dx/set_dy several times, since the distortion will just add up. 
My use case is the following: iterations of: (1) set_dx (2) calibrate (3) measure offset -> and repeat at (1) until measure is good. In principle I can use a new detector object everytime, but I think multiple calls of set_dx should not be added up.

Best wishes,
Clemens

kif:
On Wed, 15 Jul 2020 07:12:45 -0700
Clemens Prescher <notifications@github.com> wrote:

> Hi!
> 
> thank you for your fast response and change. I think it looks good. However, I think the current implementation has an issue if you call set_dx/set_dy several times, since the distortion will just add up. 
> My use case is the following: iterations of: (1) set_dx (2) calibrate (3) measure offset -> and repeat at (1) until measure is good. In principle I can use a new detector object everytime, but I think multiple calls of set_dx should not be added up.

OK, not simple as one needs to reset x and y independently ...
I need to think about it before rushing garbage code.

kif:
Could you have a try for this: it apparently resets the distortion properly and independently on the two axis.

kif:
TODO: provide some tests ...

kif:
now it is ready for merge ...

CPrescher:
Thank you. 
I can agree that it works from my side.

Issue #1372, opened 2020-07-06
Data exported from `calib2`


kif:
It would be great if units would be available in the header of the CSV file.

kif:
I take it ... There was a TODO in the code :)

PR #1380, opened 2020-07-15
display/save radial units

kif:


kif:
close #1372

vallsv:
Is that true? The unit can be changed no?

kif:
Which part are you talking about ?

kif:
I just checked, the units are updated when the user modifies the output space.

PR #1386, opened 2020-07-20
1385 detector distortion identity

### LINKED ISSUE #1385: Double check the CSR matrix for a distortion-less detector is actually identity
There are apparently too many element in the sparse matrix to be derived from identity. 

This could save a couple of seconds in the processing of ID02

kif:
close #1385


kif:
Should impact positively the processing of ID02: probably a second won per run !

PR #1390, opened 2020-08-04
Implement simple peak-picker

kif:
This implements a simple, naive, peak-finder based on the the mean/std in the region of the peak.
There is also a criteria for the noise and for symmetry of the peak.

Execution times are in the range 1-7ms for Eiger 4M images on Kepler depending on the various thresholds.


kif:
On Volta (Tesla V100 on power9), the execution time measured from python is 2ms / Eiger4M frame with:
* patch = 7x7
* threshold = 3 sigma
* noise = 1
* radius =1

kif:
close #1355 as merged here ...

kif:
CI failing due to #1391

PR #1398, opened 2020-09-07
Define Eiger2 with CdTe sensor

### LINKED ISSUE #1396: Provide masks for Dectris Eiger2-CdTe
Half a dozen such detectors are arriving at ESRF

kif:
close #1396


kif:
tested, apparently it works

PR #1400, opened 2020-09-21
Typo and pylab fix in integration_with_python notebook

loichuder:
Typo fixes and I removed the cell `%pylab nbagg`: I got `IPython is not defined` error with it while it is working fine without it. 

I am confident it should work without `%pylab nbagg` but it would be nice to test on another setup than mine.


kif:
I guess you are testing under jupyterlab and not the notebook ...

loichuder:
Oooh, true. Hang on, let me do a round of checks on the notebook

loichuder:
Actually, it works fine in the notebook as well without `%pylab nbagg`.

PR #1401, opened 2020-09-24
Corrected typos in `average` files.

loichuder:
Fixes a bug when using `quantiles` parameters without `quantiles` methods.

### DIFF
--- a/pyFAI/app/average.py
+++ b/pyFAI/app/average.py
@@ -54,7 +54,7 @@ from pyFAI import average
 
 
 def parse_algorithms(options):
-    """Return a list of initilized algorithms from the command line"""
+    """Return a list of initialized algorithms from the command line"""
 
     methods = options.method.split(",")
     methods = set(methods) - set([""])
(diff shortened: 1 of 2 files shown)


pyFAI/app/average.py
```diff
@@ -69,8 +69,8 @@ def parse_algorithms(options):
             q2 = 1.0 - q1
         quantiles = (float(q1), float(q2))
 
-        if "quanties" not in methods:
-            logger.warning("Add quantiles to the set of methods as quantiles parameters is defined.")
+        if "quantiles" not in methods:
```
loichuder, line 72: This should fix a bug when `quantiles` was not in methods but when parameters were set. Quite unlikely to occur, I guess.

pyFAI/average.py
```diff
@@ -25,12 +25,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-"""
-
-Utilities, mainly for image treatment
-
-"""
-
```
loichuder, line 33: Removed this as I assumed it was copy-paste leftover.

kif:
LGTM, thanks for the fixes

PR #1406, opened 2020-10-02
1405 sparsify

### LINKED ISSUE #1405: Single crystal diffraction data sparsification tool
PyFAI has tools to perform the separation of Bragg peaks and background and save those sparse data in HDF5.
Implement now a script sparsify-bragg which saves those data.

My be charged to ID11

kif:
Allows the (lossy) compression of single crystal datasets.
By default, the compression ratio is about 100x,

The background is recorded assuming a polar symmetry.
The signal is sigma-clipped for 5 times (default threshold: 5 sigmas)
Then peaks are recorded (default threshold: 3 sigma).

Data are packed in HDF5 and provide a script to re-densify the data.

Close #1405

### DIFF
--- a/pyFAI/app/sparsify.py
+++ b/pyFAI/app/sparsify.py
@@ -42,7 +42,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "02/10/2020"
+__date__ = "16/10/2020"
 __status__ = "status"
 
 import os
@@ -64,7 +64,7 @@ from .. import version, load
 from ..units import R_M
 from silx.opencl import ocl
 if ocl is None:
-    raise RuntimeError("Sparsfy requires a valid OpenCL stack to be installed")
+    logger.error("Sparsfy requires a valid OpenCL stack to be installed")
 from ..opencl.peak_finder import OCL_PeakFinder
 from ..utils.shell import ProgressBar
 from ..io.sparse_frame import save_sparse
--- a/pyFAI/app/sparsify.py
+++ b/pyFAI/app/sparsify.py
@@ -1,0 +1,270 @@
+#!/usr/bin/env python3
+# coding: utf-8
+#
+#    Project: Fast Azimuthal integration
+#             https://github.com/silx-kit/pyFAI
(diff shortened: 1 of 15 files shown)


kif:
Some figures on my Corei7 2600 + Nvidia GTX750:
1800 frames from Eiger 1M (Lyzozym Dectris dataset)
Original size: 2.2G 
Packed in 29M.
Compression time: 12s
About 5000 pixel registered per frame.

jonwright:
Looks promising - I am afraid I might not get time to try to run it for a week or two due to the beamline.

There is an implementation for the peak saving here: /data/id11/jon/38keVLaB6/segment1.ipynb
(About 5-6 hours to do 3.2M frames for the peaks only). That hooks up to my peaksearch code in the same folder, but I think your sparse format can also plug in easily to that code (just converting flat index to row/col).

There was a slight change I wanted for the script to read the sparse pixels here (to get one frame at a time) in here:
/data/id11/jon/JKsparse/JKSparse.ipynb

One of the recurring problems is to know which motors/counters are needed to go with the frames and which scans to input and how to tell the program what you want it to do (not specific to this case, but it will come up on real data).

kif:
> Looks promising - I am afraid I might not get time to try to run it for a week or two due to the beamline.

It's up to you.

> There is an implementation for the peak saving here: /data/id11/jon/38keVLaB6/segment1.ipynb
> (About 5-6 hours to do 3.2M frames for the peaks only). That hooks up to my peaksearch code in the same folder, but I think your sparse format can also plug in easily to that code (just converting flat index to row/col).

This implementation is based on GPU and is likely to be much faster. The "azimuthal" error model is slower but not much slower (i.e. 14s instead of 12s, many more peaks found like 10k instead of 5k which would be enough to justify the difference of speed).

> There was a slight change I wanted for the script to read the sparse pixels here (to get one frame at a time) in here:
> /data/id11/jon/JKsparse/JKSparse.ipynb

I'll have a look. Maybe it makes sense to provide the text of a function to extract one frame rather than extracting eveything.

> One of the recurring problems is to know which motors/counters are needed to go with the frames and which scans to input and how to tell the program what you want it to do (not specific to this case, but it will come up on real data).

Yop, but this is a spec/bliss problem. It cannot be addressed here. A side-car file is probably the best and also the cheapest.

Jerome

kif:
Damn, I need to make those tests pass without opencl !

Issue #1418, opened 2020-10-28
Issues about the documentation


kif:
We discussed with @loichuder @takluyver and found those issues in the documentation:
- [x] Start the documentation with a sketch of the transformation performed within pyFAI
- [x] Remove the video from the main page (-> to the publication sections ?)
- [x] Switch the default theme to the RTD one (at least it looks consistent between RTD and silx.org)
- [x] Put forward the scheme with the geometry and what the poni-file is (i.e. the file containing the geometry)
- [x] Definition of the poni-file in the glossary, link to the proper section
- [x] Restructure `cookbooks` and include them within `tutorials`, maybe as `HowTos`
- [x] Provide video for the cookbooks explaining how to use the GUI applications (<5mn)
- [x] Expose the 5 layers of pyFAI and link the API doc from there.  
- [x] Label explicitly deprecated scripts.
- [x] Do not document apps in the API doc, but Nexus is missing for example.
- [x] Link the documentation from the readme file (use badges?)

kif:
Attempt for the first sketch:
![overview](https://user-images.githubusercontent.com/1018880/97465636-8849b580-1942-11eb-9239-64652b9bbd9f.png)

loichuder:
Looks good ! I would perhaps make the arrows start from the 2D diffraction pattern (above PyFAI).

takluyver:
Thoughts on the image:

- I agree with Loïc - it could be clearer that the top image is input.
- It's not immediately obvious what the 3D image is showing. Peering closely, I think I can see that the detector image is tilted within the axes box. But I'm not sure how I'd make use of that transformation apart from looking at a 3D view. Maybe there's a better way to illustrate what's going on? Or maybe it's better to leave this out?
- It's not very obvious that the bottom image is a transformation of the top one. Can we make the colour scales match closer? And maybe lose the overlaid green lines, so the lines in the image itself are more visible?
- I think the Wordart style 3D text looks kind of 1990s. But if you're happy with that look, don't let me put you off. :wink:

kif:
For the 3D setup view, I am neither happy ... maybe we should find something else...
https://github.com/silx-kit/pyFAI/blob/c29764b040ae8bb1e6b04c3697713d17ce8f18b7/doc/source/img/overview.png?raw=true
About the Wordart ... sorry, I am an old guy, back from that time :-þ

kif:
Could you please have a look at the upgraded documentation ?
[...]

PR #1419, opened 2020-10-28
Doc

kif:
related to #1418 
new image for the main page of the doc.


kif:
Not yet included into the doc.

PR #1428, opened 2020-11-05
1424 export mask as rectangles

kif:
This is a minor improvement with proper memset performed on memory-views


kif:
On the detector from ID11 it decreases the number of rectangles from 480 to 370 ...

PR #1432, opened 2020-11-30
Fixed typo in sphinx config

t20100:
This PR aims at fixing debian packaging which looks to be broken by PR #1420 (see https://gitlab.esrf.fr/silx/bob/pyfai/-/pipelines/38048):
```
File "conf.py", line 75, in <module>
ImportError: No module named 'sphinx_rtd_theme'
```

If the goal was to use RTD sphinx theme by default, then it needs to be added to debian control build environment.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -71,7 +71,7 @@ extensions = [
 if sphinx.__version__ < "1.4":
     extensions.append('sphinx.ext.pngmath')
 
-if not on_rtd:
+if on_rtd:
     import sphinx_rtd_theme
 
     extensions.append('sphinx_rtd_theme')


kif:
The idea is rather to create a dependency on RTD...
https://github.com/silx-kit/pyFAI/pull/1431/commits/2bd7e887f390769722ec2498fbb78a438e1284a5

loichuder:
The aim of this condition was rather to set the theme to `sphinx_rtd_theme` when *not* building on Read The Docs.  Read The Docs indeed uses its own theme anyway.

Sorry it broke everything :pray:

t20100:
OK, I understand now, on RTD "default" is "rtd-theme"... that's why it sounds weird in the code.... Then https://github.com/silx-kit/pyFAI/commit/2bd7e887f390769722ec2498fbb78a438e1284a5 fixes it.

PR #1445, opened 2021-01-12
Migration

kif:
Big refactoring to allow the generalization of new-generation rebinning engines.


kif:
This Tiny PR replaces basically `integrate1d_legacy` with `integrate1d_ng`
The same remains to be done for the 2D version. But it is unlikely I find the time before next week.

Issue #1448, opened 2021-01-14
[GPU] OpenCL integrator should carefully chose the block-size for optimal performances


kif:
silx provides now a min-workgroup-size attribute.

PR #1452, opened 2021-01-20
Dynamic worgroup

kif:
The core idea is to probe each kernel in order to work with the recommended (optimal ?) workgroup size.

Tests have also been refactored. 
Broken configuration are also avoided.


kif:
close #1448

Issue #1461, opened 2021-01-28
Debian packaging of 0.20.0


picca:
Hello, Jerome, I will put here all the issue found during the packaging.

during the build_man processing

```
ERROR:pyFAI.app.sparsify:Sparsfy requires a valid OpenCL stack to be installed
	pybuild --build -i python{version}-dbg -p 3.9 -s custom -p 3.9 "--build-args=PYTHONPATH={build_dir} xvfb-run -a --server-args=\"-screen 0 1024x768x24\" {interpreter} setup.py build_man"
I: pybuild base:232: PYTHONPATH=/builds/science-team/pyfai/debian/output/source_dir/.pybuild/cpython3_3.9_dbg_pyfai/build xvfb-run -a --server-args="-screen 0 1024x768x24" python3.9-dbg setup.py build_man
```

There is no OpenCL available during the build so the manual page can not generated :)

picca:
Here a bunch of warning due to unclosed files

```
INFO:pyFAI.setup:Build man for entry-point target 'check_calib'
/builds/science-team/pyfai/debian/output/source_dir/setup.py:349: ResourceWarning: unclosed file <_io.BufferedReader name=3>
  synopsis = self.get_synopsis(module_name, env)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
INFO:pyFAI.setup:Build man for entry-point target 'detector2nexus'
/builds/science-team/pyfai/debian/output/source_dir/setup.py:349: ResourceWarning: unclosed file <_io.BufferedReader name=3>
  synopsis = self.get_synopsis(module_name, env)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
INFO:pyFAI.setup:Build man for entry-point target 'diff_map'
/builds/science-team/pyfai/debian/output/source_dir/setup.py:349: ResourceWarning: unclosed file <_io.BufferedReader name=3>
  synopsis = self.get_synopsis(module_name, env)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
INFO:pyFAI.setup:Build man for entry-point target 'diff_tomo'
/builds/science-team/pyfai/debian/output/source_dir/setup.py:349: ResourceWarning: unclosed file <_io.BufferedReader name=3>
  synopsis = self.get_synopsis(module_name, env)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
INFO:pyFAI.setup:Build man for entry-point target 'eiger-mask'
[...]

PR #1462, opened 2021-01-28
Fix some bugs reported during debian packaging

kif:
fix numerous warnings when building manual pages


kif:
close #1461

kif:
Hi Fred, does this help on your i386 ?

kif:
If this is enough for Debian ... let's merge it.

PR #1474, opened 2021-03-02
Typo in tooltip

kif:


kif:
Typo reported by Alba synchrotron.

vallsv:
It could be good to use a `QDial` here https://doc.qt.io/qt-5/qdial.html in order to have a better look and feel that's about the direction.

PR #1476, opened 2021-03-02
Fixed a few typos in `units` module doc

loichuder:


pyFAI/units.py
```diff
@@ -57,15 +57,15 @@
...
-in Angstrom.KeV. It is approximativly equal to:
- 
-pyFAI reference 12.398419292004204
-scipy v1.3.1:   12.398419739640717
-scipy-1.4.0rc1: 12.398419843320026
+in Angstrom.KeV. It is approximatively equal to:
+
+- pyFAI reference: 12.398419292004204
+- scipy v1.3.1:   12.398419739640717
+- scipy-1.4.0rc1: 12.398419843320026
```
loichuder, line 64: I find this section misleading. What is `pyFAI reference` ? What value can the user expect ?

pyFAI/units.py
```diff
@@ -57,15 +57,15 @@
...
 """
 
 CONST_q = scipy.constants.e
-"""One electron-volt is equal to 1.602176565⋅10-19 joules"""
+"""One electron-volt is equal to 1.602176634⋅10-19 joules"""
```
loichuder, line 68: Changed in [scipy 1.4.0](https://docs.scipy.org/doc/scipy/reference/release.1.4.0.html#scipy-constants-improvements)

pyFAI/units.py
kif: This constant used to be hard-coded. Now it relies on scipy which is probably better ... but when scipy changes its value, some tests needs to be adapted. It is good to keep some track of those constants when some tests are randomly failing.

pyFAI/units.py
kif: Thanks for updating it ...

PR #1480, opened 2021-03-08
Wire the export button

### LINKED ISSUE #1479: pyFAI-intergrate, export button does not work
Sometimes it would be handy to export some modified geometry as PONI-file

kif:
Valentin, I am sure you would do this differently ...
maybe you could have an eye on it ?

close #1479


vallsv:
LGTM

PR #1497, opened 2021-03-30
expose workgroup selection tool

kif:


kif:
tested in prod ... this works

PR #1501, opened 2021-04-06
Distance was not read-back !

### LINKED ISSUE #1500: bug in reading back geometry from JSON string
it is "dist", not "distance". 
The code is not self consistent.

kif:
close #1500


kif:
Does it deserve a non-regression test ?

kif:
not sure ... the issue is fixed where it was found

PR #1511, opened 2021-05-12
Polarization for sparsification

### LINKED ISSUE #1508: Take into account polarization and other normaliation for Sparsify-bragg

kif:
close #1508


kif:
The bug in RTD is something different: https://github.com/spatialaudio/nbsphinx/issues/563

Issue #1515, opened 2021-05-20
[OpenCL] Replace Kahan with double-word library


kif:
The double-word library has been certified, it is integrated into **silx** now and it is up to 2x faster ...
It is probably the right time to clean-up some OpenCL code as well.

PR #1516, opened 2021-05-31
Use the double-word OpenCL library from silx

kif:
About performances ...
* On CPU drivers Intel and PoCL: sllightly slower. An actual double-precision version would be better
* On Nvidia GPU: No noticeable differences were seen.
* One AMD GPU: Performances looks much better, sometimes 30% better !


kif:
close #1515

kif:
Merge only after the release of silx _and_ set the minimal version for silx !

kif:
since silx 0.15 is out, let's merge it !

PR #1522, opened 2021-06-02
1521 fixed param

### LINKED ISSUE #1521: [calib2] Fixed parameters are reset when new control points are added
Example: 
* set rot3=90° and lock it,
* add new points in the peak-picking tab
* Come back to the fit-tab, rot3 is again at 0° !

kif:
close #1521


kif:
@vallsv do you have a bit of time to look at this ?

vallsv:
Could we take a look together on of this days?

kif:
I am onsite Thursday ... Nils Blanc also reported the bug

Issue #1537, opened 2021-07-29
windows cp39 wheel requires numpy>=1.21.1


ericpre:
If I understand correctly
https://github.com/silx-kit/pyFAI/blob/97ddf891bcbf4acb921036ae4f3be2a5cb48c857/setup.py#L982-L994
it seems that this wheel (`pyFAI-0.20.0-cp39-cp39-win_amd64.whl`) has been build against numpy 1.21.1, which requires numpy 1.21.1 at runtime.

For example:
```
pip install pyfai==0.20 numpy==1.21.0
```
Give the following errors:
```
The conflict is caused by:
    The user requested numpy==1.21.0
    pyfai 0.20.0 depends on numpy>=1.21.1
```

Would it be possible to build again the wheel with an older numpy version?

ericpre:
If compiler are available on the system, a workaround to the issue with the current package on pypi is to use: 
```
pip install pyfai==0.20 --no-binary pyfai --no-build-isolation
```

PR #1538, opened 2021-07-31
Fix numpy build dependency 

ericpre:
Fixes https://github.com/silx-kit/pyFAI/issues/1537 by using https://github.com/scipy/oldest-supported-numpy as build dependency.

I am suggesting this bugfix but I didn't check if this is fully working - I don't have easy access to a windows system with a development environment setup.


kif:
LGTM, thanks for the help.

PR #1541, opened 2021-08-20
Parallax effect correction

### LINKED ISSUE #1529: Correct pixel position for parallax effect
If one considers that the absorption in the sensor material follows Beer-Lambert's law, the residual beam can be expressed as a first order exponential decay: `exp(-µx)`. 
Let `α` be the incidence angle. The inclining of the beam shifts the peak position by `sin(α)/µ` for an *infinitely* thick detector. 
For a detector of finite thickness, this peak-position offset can be calculated from a numerical convolution with the beam profile. The ideal would be to expose this correction directly as a function of `sin(α)`.

This can be divided in sub-tasks:
* [x] Expose `sin(alpha)` in the geometry. Since `cos(α)` already exists, the most efficient is to calculate them both.
* [x] Implement a `Parallax` class with a sensor thickness, the depth of penetration  (µ), and a beam profile ...
  This class is responsible for providing an offset (in meter) for a given sin(alpha).
* [x] Correct the coordinates in xyz of the pixel position for the parallax effect
* [x] validate with some tests.

PS: Parallax has also an effect on the apparent efficiency which can be addressed with the absorption keyword in azimuthal integration.

kif:
Close #1529


kif:
Despite I am not really happy with this PR, I merge it into 0.21 so that it is available and will re-work on it later.

PR #1550, opened 2021-09-07
Improve multi-geometry generation

kif:
correct one bug and expose many constructor parameters of MultiGeometry.


kif:
close #1551

PR #1563, opened 2021-09-24
Peakfinder8

kif:
This is an attempt to expose peakfinder8 algorithm for fast integration of serial crystallography frames.
* [x] Make code work on all platforms
* [x] Benchmark on various platforms
* [x] Investigate the huge shared memory requirement
* [x] Provide some non regression tests


kif:
close #1563

kif:
Power9 + Tesla V100 is <5ms for 6 Mpix image, 50X faster than C++ implementation.
Xeron + RTX A5000 is <6ms for same image. 20x faster than C++

PR #1568, opened 2021-10-15
Rework peak-picking in matplotlib

kif:
Two ideas:
- fix the calibration with pyFAI-calib under matplotlib 3.4+
- Prepare a similar calibration engine under jupyter notebooks or jupyterlab based on matplotlib


kif:
@satarsa could you please test this branch ?
Since I rewrote many parts, I would not be surprised if some things got broken (and there are no tests on the GUI)

Issue #1577, opened 2021-11-13
improve tab order in GUIs


kremeyer:
Whenever I use the GUIs of pyFAI, I notice that in some dialogs the order in which input boxes are selected when cycling with the tab key is rather arbitrary. I suppose that the tab cycling should start at the top left and work it's way down to the bottom right.

### Example
Before:
<img width="258" alt="old" src="https://user-images.githubusercontent.com/40638974/141658007-a79015cc-f9a7-490c-9b6d-21b398aee34d.png">

After:
<img width="262" alt="new" src="https://user-images.githubusercontent.com/40638974/141658020-4bcf62be-de4b-4e2a-bf43-242d540bfd8c.png">

I already prepared this for all .ui files in https://github.com/kremeyer/pyFAI/commit/afc5d34b5eddb567067d425f09d213f2832ef9de

vallsv:
CLosed by #1578

PR #1578, opened 2021-11-14
fixed tab order to be from left to right and from top to bottom

kremeyer:
resolving tab order as described in issue #1577


vallsv:
Sounds good as far as i can see.

Thanks a lot for the patch.

Issue #1444, opened 2021-01-11
Missing integrators for replacing integrate_legacy with integrate_ng


kif:
[ ] IntegrationMethod(1d int, bbox split, histogram, cython)
[ ] IntegrationMethod(1d int, full split, histogram, cython)

kif:
To be validated but the matrix is almost complete ...

PR #1580, opened 2021-11-18
1444 2d inherit from csr

kif:
Perform 2D integration using the CsrIntegrator parent class.
Minor adaptation there


kif:
This is needed for #1444 but there is a long way to go !

PR #1584, opened 2021-11-19
Rework all LUT integrators

kif:
* 2d integrator inherit from parent class and the integration methods
* Use sparse builder everywhere (hence switch to C++ code)
* implement full pixel splitting both for 1D and 2D.


kif:
Fixing the test is left in #1589. Merge for now.

PR #1599, opened 2021-12-10
Fix fullsplit tests

kif:
Global clean up of all Cython code


kif:
Again ... 200 lignes of Cython saved.

PR #1603, opened 2021-12-14
fix #1602

### LINKED ISSUE #1602: problem with integrate_radial with csr_ocl
Hi, 
I tried to run a radial integration with the csl_ocl method and I get the following error:
```
  File "<ipython-input-32-af49e0b0ffe6>", line 1, in <module>
    res = ai.integrate_radial(raw, 36, method='csr_ocl')

  File "C:\Program_writable\Anaconda3\envs\pydidas\lib\site-packages\pyFAI\azimuthalIntegrator.py", line 1617, in integrate_radial
    intensity[count == 0] = empty

TypeError: 'numpy.float32' object does not support item assignment
```
Using for example csl works fine. I checked the type and shape of the sum_ and count variables (lines 1612/1613) and they are 
sum with CSR_OCL: ```<class 'numpy.float32'>```
count with CSR_OCL: ```<class 'numpy.float32'>```
sum with CSR: ```<class 'numpy.ndarray'>```
count with CSR: ```<class 'numpy.ndarray'>```

and when I check the shapes of the res.sum and res.count from the integrate2d, they are
with CSR:
res.count.shape (36, 100)
res.sum.shape (36, 100)

and with CSR_OCL
res_ocl.count.shape (3600,)
res_ocl.sum.shape (3600,)

The same is true for LUT and LUT_OCL. The sum over axis -1 then yields a float for the _OCL methods which raises the TypeError.

I would assume this behaviour is a bug? I could reproduce the problem in Windows and Linux. I have looked through the list of the open issues and (judging by name) think this issue has not yet been reported.

kif:
close #1602
Tests are still missing

### DIFF
--- a/pyFAI/azimuthalIntegrator.py
+++ b/pyFAI/azimuthalIntegrator.py
@@ -1637,31 +1637,48 @@ class AzimuthalIntegrator(Geometry):
         :rtype: Integrate1dResult
         """
         unit = units.to_unit(unit, type_=units.AZIMUTHAL_UNITS)
-        res = self.integrate2d(data, npt_rad, npt,
-                               correctSolidAngle=correctSolidAngle,
-                               mask=mask, dummy=dummy, delta_dummy=delta_dummy,
-                               polarization_factor=polarization_factor,
-                               dark=dark, flat=flat, method=method,
-                               normalization_factor=normalization_factor,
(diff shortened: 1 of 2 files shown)


kif:
Improve test coverage here ...

PR #1604, opened 2021-12-15
Jupyterlab calib

kif:


kif:
get inspired by
https://github.com/matplotlib/ipympl/issues/270

PR #1605, opened 2021-12-21
Testimages

kif:
Restore the `python setup.py testimages` command


kif:
manually tested with and without this PR from silx: https://github.com/silx-kit/silx/pull/3580

PR #1630, opened 2022-01-28
Propagate also uncertainties with OpenCL/Python

kif:


kif:
So plenty of bugs in the wiring between AzimuthalIntegrator and the underlying rebinning engines

PR #1641, opened 2022-02-10
API: restore back compatibility in histoBBox1d and histoBBox2d

tacaswell:
The signatures of a number of functions were changed in #1528.  This restores
the ability to pass by keyword pos0Range and pos1Range in histoBBox1d adn
histoBBox2D.

If both the old and new parameters are passed, `ValueError` is raised, if only
the old keyword is passed a warning is issued.  If the value is passed
as a positional the users is unaware of this change.

-----

This was brought to my attention because it broke @gfreychet . 

If you want to rename a lot of keywords, it may be worth vendoring a version of a decorator we have in Matplotlib that handles all of the "only pass one" logic.

### DIFF
--- a/pyFAI/ext/splitBBox.pyx
+++ b/pyFAI/ext/splitBBox.pyx
@@ -42,6 +42,8 @@ __license__ = "MIT"
 
 include "regrid_common.pxi"
 import logging
+import warnings
+
 logger = logging.getLogger(__name__)
 from .splitBBox_common import calc_boundaries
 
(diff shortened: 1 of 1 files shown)


kif:
LGTM

kif:
There has been a major rewrite of all integrators for 0.21 and I aligned them all with the same API (thus breaking the backwards compatibility). I did not think anybody (else than me and AzimuthalIntegrator) was using them directly ...

It would be nice to have those incompatible changes tracked and have them raising a warning.  I will make a minor release soon with those fixes + deprecation warnings.

But for now, I am off for ⛷️ , don't worry if I am inactive in the coming weeks.

PR #1650, opened 2022-03-11
Avoid empty bins in sigma-clipping

kif:
Investigation on empty bins when performing sigma-clipping with Poisson error model


kif:
Finally this is a minor bug-fix on CI
close #1651

PR #1682, opened 2022-04-22
added splitters

iltommi:
Dear developers, 
I replaced the horizontal layout with panel splitting so that the right part can be stretched out (reducing the image size) in order to better accomodate the widgets. It's particularly useful in the geometry panel where the values of the PONIs and rotations are quite large (I was missing the 10**-9 exponent that was on the rightmost part of "rotation3"...)

here is a screenshot of the result (you can see the draggable dot in the middle between the two panels):
Screenshot 2022-04-22 at 16 35 22

The diff is huge just because the ui gets scrambled out when removing the outermost layout (I used the qt-designer).

Thanks for this great software!

### DIFF
--- a/pyFAI/resources/gui/calibration-experiment.ui
+++ b/pyFAI/resources/gui/calibration-experiment.ui
@@ -13,402 +13,402 @@
   <property name="windowTitle">
    <string>Experiment settings</string>
   </property>
-  <layout class="QHBoxLayout" name="horizontalLayout" stretch="1,0">
-   <property name="margin">
-    <number>0</number>
-   </property>
+  <layout class="QHBoxLayout" name="horizontalLayout">
    <item>
-    <widget class="QFrame" name="_imageHolder">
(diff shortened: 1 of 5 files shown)


kif:
@vallsv could you please have a look ? thanks

kif:
Hi,
I just checked the code and the automatic compression of the left tab (navigation tab) stopped working. 
More work may be needed.
Jerome

kif:
Tried some basic fiddling in XML file without success. @vallsv , your expertise will be needed.

vallsv:
Hi, sounds good to me.

In the PR, i suggest you to set `collapsible` property to false.

But, if one of this long numbers is not completely displayed.
you still have to do an action manually, which is probably
not the safest way to handle the problem.

So i think it would be better to set the min size of the line edit based on the content, instead of using splitter.

But it's up to you.
Regards,

PR #1684, opened 2022-05-02
Ensures the LineEdit widget is large enough to host all digits

kif:
prevents data from being elided

Alternative to #1682


kif:
@iltommi would this fix the issue you mention in #1682 ?

PR #1702, opened 2022-06-10
replace do_poisson (bool) with error_model (str)

### LINKED ISSUE #1701: worker does not allow for azimuthal error model
No matter what I try, the worker does not seem to accept error models otther than poisson

```python
from pyFAI.worker import Worker

worker = Worker(extra_options={"error_model", "azimuthal"})
worker.do_azimuthal_error = True
worker.set_config({"error_model", "azimuthal", "do_azimuthal_error":True})
worker.error_model = "azimuthal"
```

woutdenolf:
Closes #1701

### DIFF
--- a/plugins/Lima/AzimuthalIntegration.py
+++ b/plugins/Lima/AzimuthalIntegration.py
@@ -61,7 +61,7 @@ class SinkPyFAI(Core.Processlib.SinkTaskBase):
         self.mask_image = None
         self.subdir = ""
         self.extension = None
-        self.do_poisson = None
+        self.error_model = None
 #        self.do
         try:
             self.shapeIn = (camera.getFrameDim.getHeight(), camera.getFrameDim.getWidth())
@@ -168,10 +168,8 @@ class SinkPyFAI(Core.Processlib.SinkTaskBase):
                 kwarg["filename"] += self.extension
             else:
                 kwarg["filename"] += ".xy"
-        if self.do_poisson:
-            kwarg["error_model"] = "poisson"
-        else:
-            kwarg["error_model"] = "None"
+        if self.error_model:
+            kwarg["error_model"] = self.error_model
 
         try:
             if self.do_2D():
@@ -274,7 +272,7 @@ class SinkPyFAI(Core.Processlib.SinkTaskBase):
         if config.get("rad_pt"):
             self.nbpt_rad = int(config.get("rad_pt"))
         self.unit = pyFAI.units.to_unit(config.get("unit", pyFAI.units.TTH_DEG))
-        self.do_poisson = config.get("do_poisson")
+        self.error_model = config.get("error_model")
         if config.get("do_polarization"):
             self.polarization = config.get("polarization")
         else:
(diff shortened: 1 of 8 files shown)


kif:
The PR looks OK but it won't fix your issue since the lower level part is not wired (yet) ...

kif:
Maybe it is not that difficult to replace "integrate1d" with "sigmaclip_ng" in the method `update_processor` from the processor. Other parameters should be the same. Mind to enforce "cycle=0" to prevent actual clipping.

woutdenolf:
It does fix my issue.

```python
worker_options = {
    "integrator_name": "sigma_clip_ng",
    "extra_options": {
        "max_iter": 3,
        "thres": 3,
    },
}
integration_options = {
    "method": "bbox_csr_opencl",
    "unit": unit,
    "nbpt_rad": nbpt_rad,
    "polarization_factor": polarization_factor,
    "error_model": "azimuthal",
}
```

The first is used to instantiate the worker, the second to configure the worker.

Of course some parameter combinations may raise an error because the lower level doesn't support it (e.g. integrate1d in combination with azimuthal error model). We could assert these forbidden combinations but you already get the exception from the lower level I presume, so I don't see any urgency for that.

woutdenolf:
> Mind to enforce "cycle=0" to prevent actual clipping.

With `cycle` I presume you mean `max_iter`. Why would I want to prevent actual clipping? Isn't that the whole point clipping?

What I want (or the client wants) is azimuthal integration while removing spots. 

So I thought sigma clipping with an `azimuthal` error model with `thres=3` would reject outliers alone the ring with an intensity of 3*sigma where sigma the standard deviation along the ring (`azimuthal` error model). I don't know what `max_iter` does but I guess it should >0 to have it remove the outliers. Perhaps I'm missing something?

kif:
Sorry, I thought you were doing azimuthal integration, if you want to perform pixel-rejection, it is perfectly adapted.
`cycle` and `max_iter` are the same, `cycle` is the name at the lower level interface (OpenCL) and correspond to the number of iteration of clipping.
Few comments on this since I spent a lot of time on the topic this year:
* cycle can be as low as 3 without major degradation, at least this is what I checked on MX images
* the threshold can be set to zero to tell pyFAI to use a fail-safe criterion (Chauvenet) depending on the size of the ensemble at a given bin. the actual value varies from 2.5 to 3.5.
* you can ask pyFAI to perform the sigma-clipping with azimuthal error model and export the uncertainties based on Poisson statistics, this tends to save 1 iteration in the sima-clipping (not yet demonstrated).
Most of this is written in a manuscript I will submit by this summer ... maybe you will be interested in having a look at it.

woutdenolf:
> Sorry, I thought you were doing azimuthal integration

Well I am doing azimuthal integration with outlier rejection. At least that's what I think I'm doing with `sigma_clip_ng`.

The other points go right over my head sorry :-)

woutdenolf:
> you can ask pyFAI to perform the sigma-clipping with azimuthal error model and export the uncertainties based on Poisson statistics

Aha so how do I do that? I only see one `error_model` parameter.

kif:
the paper explaining the method is under development at
https://github.com/kif/publi_serial
you should be able to compile it with latex

woutdenolf:
Thanks, I'll have a look.

PR #1706, opened 2022-06-24
1703 sigma clip

kif:
Fix issue with normalization when playing with sigma-clipping.
- Update formula of `std` and `sem` in all core engines
  * [x] Python 
  * [ ] Cython
  * [ ] OpenCL
- [ ] General wiring in upper level (azimuthal integrator)
- [x] Non regression test


kif:
Awaiting PR from 
https://github.com/kif/pyFAI/pull/12

PR #1727, opened 2022-08-29
add support for pilatus-900kw detector

rtellez700:


kif:
Hi,

Thanks for the PR. The shape of this detector puzzles me since the `Pilatus 900kw CdTe` has a shape of (195, 4439) and your of (619,1475). 
According to the Pilatus module definition, it could be an assembly of 3x3 modules. In this case, the best would be to call this detector `Pilatus 900k` (i.e. not wide).

Can you provide also some hints where this detector is in use, since it looks like to have an unusual form-factor and not many of them will be available world-wide.

Cheers, 
Jerome

rtellez700:
Hi Jerome,

I also found it odd that it had a different shape than the CdTe version. I'm currently a user of the NSLS-II 12-ID beamline and they recently installed this new detector (https://www.bnl.gov/nsls2/beamlines/beamline.php?r=12-ID). They currently have an implementation on their smi-analysis (https://github.com/NSLS-II-SMI/smi-analysis) package  but has some bugs I'm trying to fix.

Thanks,
Rodrigo

PR #1729, opened 2022-08-30
skip tests that are unreliable on 32-bit platforms by default

mwhudson:
These tests now fail all the time on armhf in Ubuntu, and I could pass lowmem to skip them I guess but it seems better to skip these tests by default and avoid losing the coverage of the other lowmem tests (which all seem to pass fine).


kif:
LGTM. Thanks for the contribution.

Do not formalize about RTD failing since it is addressed elsewhere.

PR #1733, opened 2022-09-08
Implement a CSC integrator

### LINKED ISSUE #1717: Single threaded 1D integration : performance glitch?
I am seeing something like 10 fps per core for our Eiger 4M images, which seems to be in line with the pyFAI-benchmark plots in the docs. 

For some reason I timed a [fast_histogram](https://github.com/astrofrog/fast-histogram).histogram1d and this seems to offer 10 ms, which would be 100 fps, but no pixel splitting etc. That number motivated some messing about to take your CSR matrix try to tune it to reduce the matrix data size and then run the dot product via numba. So far it reached 18 ms, or about 50 fps for the dot product. These seem to be worth making into patches if they are getting the numbers right - what do you think?

The CSR python code in pyFAI is giving: "CPU times: user 119 ms, sys: 32.4 ms, total: 151 ms", which is consistent with doing two dot products with scipy.sparse.csr.dot at 59 ms each? The second product seems like it could be cached for processing a series of frames, but maybe I have missed something that changes. Do you know what the system time is used for ?

Some code should be attached: [SingleCore1D.md](https://github.com/silx-kit/pyFAI/files/9073925/SingleCore1D.md)

kif:
close #1717


kif:
Here is quick benchmark on a dual Epyc7343 (2x16core+hyperthreading), execution time in seconds for the integration of 1000 frames as function of the number of threads:
```
1 12.713942694477737
2 6.027193841524422
4 3.0942815160378814
8 1.430577140301466
16 0.7645775210112333
32 0.6310557564720511
64 0.4246820630505681
```
So for 64 threads, one achieves a speed-up of 30x !

kif:
Figure_1
Up to 10Mpix images, the serial version of sparse matrix multiplication is even faster than the parallel version, ON THIS SPECIFIC PROCESSOR.

pyFAI/azimuthalIntegrator.py
```diff
@@ -1202,7 +1190,7 @@ def integrate1d_ng(self, data, npt, filename=None,
                 variance = (numpy.maximum(data, 1.0) + numpy.maximum(dark, 0.0)).astype(numpy.float32)
 
         # Prepare LUT if needed!
-        if method.algo_lower in ("csr", "lut"):
+        if method.algo_lower in ("csr", "lut", "csc"):
```
jonwright, line 1193: Could this kind of decision be inside the 'method' object?

pyFAI/ext/CSC_common.pxi
```diff
@@ -0,0 +1,308 @@
...
+    @property
+    def indices(self):
+        return numpy.asarray(self._indices)
+    @property
+    def indptr(self):
+        return numpy.asarray(self._indptr)
+
+
+    def integrate_ng(self,
+                     weights,
```
jonwright, line 103: Are these the raw pixel intensities, before the dark/flat/mask etc. For 'weights' I would think about something else.

pyFAI/ext/CSC_common.pxi
```diff
@@ -0,0 +1,308 @@
...
+    
+        with nogil:
+            #first loop for pixel in input image
+            for idx in range(self._indptr.shape[0]-1):  
+                # skip pixel if masked:
+                start = self._indptr[idx]
+                stop = self._indptr[idx+1]
+                if stop == start: # pixel contributes to no bin
+                    continue
+                is_valid = preproc_value_inplace(&value,
```
jonwright, line 228: Does this get inlined?

pyFAI/ext/CSC_common.pxi
```diff
@@ -0,0 +1,308 @@
...
+        if absorption is not None:
+            do_absorption = True
+            assert absorption.size == self.input_size, "absorption array size"
+            cabsorption = numpy.ascontiguousarray(absorption.ravel(), dtype=data_d)
+        else:
+            do_absorption = False
+    
+        with nogil:
+            #first loop for pixel in input image
+            for idx in range(self._indptr.shape[0]-1):
```
jonwright, line 222: For 1D outputs that do not fit into L1 cache, it might make sense to loop over tiles on the input image.

pyFAI/ext/CSC_common.pxi
```diff
@@ -0,0 +1,308 @@
...
+                            sum_norm_sq[bin] = omega2_A + omega2_B
+                            # omega3 = acc_norm * omega_A * omega2_B
+                            # VV_{AUb} = VV_A + ω_b^2 * (b-<A>) * (b-<AUb>)
+                            b = value.signal / value.norm
+                            delta1 = sum_sig[bin]/omega_A - b
+                            sum_sig[bin] += coef * value.signal 
+                            delta2 = sum_sig[bin] / sum_norm[bin] - b                        
+                            sum_var[bin] += omega2_B * delta1 * delta2
+                            sum_count[bin] += coef * value.count
+                    else:
```
jonwright, line 273: Are there 5 integrations being done here? Above it says "error_model=ErrorModel.NO", so I would expect some of them could be skipped?

Also, I had in mind that only `sum_sig` changes when processing a series of frames with a constant mask. When doing dynamic masking, there should be an alternative route that keeps cached sums and subtracts for the w/count of masked pixels.

pyFAI/load_integrators.py
```diff
@@ -128,6 +128,33 @@
...
                       class_funct_legacy=(py_CSR_engine.CsrIntegrator2d, py_CSR_engine.CsrIntegrator2d.integrate))
 
+try:
+    from .ext import splitBBoxCSC  # IGNORE:F0401
+except ImportError as error:
+    logger.error("Unable to import pyFAI.ext.splitBBoxCSC"
+                 " CSC based azimuthal integration: %s", error)
+    splitBBoxCSC = None
+else:
+    # Register splitBBoxCSR integrators
```
jonwright, line 138: CSR -> CSC in comment

jonwright (commented):
Feel free to merge if your CI is passing. Many of these comments reflect that I have not understood how the _ng methods are working with dynamic masking. Those question are not specific to this csc code.

pyFAI/ext/CSC_common.pxi
jonwright: Found in pyFAI/ext/regrid_common.pxi: it should be inlined. But I am surprised that norm is not being cached?
`norm = normalization_factor * flat * polarization * solidangle * absorption`. 

A related question: can the dark subtraction be done on the 1D integrated data instead of the 2D image. e.g:

`signal = sum( pixel - dark ) / sum( normalisation ) = ( sum( pixel ) - sum( dark ) ) / sum( normalisation )`

Like this, it appears it would read 6 x 2D images [ data, dark, flat, polarization, solidangle, absorbtion ] per integration] as well as a mask. For a series of images with the same geometry (etc) then I wonder if numbers can be computed by only reading the data, and then subtracting dynamically masked pixels from cached 1D sums?

pyFAI/azimuthalIntegrator.py
```diff
@@ -401,20 +406,66 @@ def setup_LUT(self, shape, npt, mask=None,
...
-                                                mask_checksum=mask_checksum,
-                                                allow_pos0_neg=False,
-                                                unit=unit,
-                                                chiDiscAtPi=self.chiDiscAtPi,
-                                                empty=empty)
+        algo = algo.upper()
+        if algo == "LUT":
+            if split == "full":
+                if int2d:
+                    return splitPixelFullLUT.HistoLUT2dFullSplit(pos,
```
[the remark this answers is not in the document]
kif: On the other hand, without this short refactoring, it would requires +200 loc of duplicated code.

pyFAI/azimuthalIntegrator.py
kif: Yes it could, it is even foreseen to be made like this but it will requires a major refactoring of all the `ai.integrate1d_ng` method.

pyFAI/ext/CSC_common.pxi
kif: Yes, those are the raw pixel intensities. The name "weights" comes from the time when intensities were used as weights for histogramming the pixel position.

pyFAI/ext/CSC_common.pxi
kif: Several question: yes, up to 6 arrays can be read for a a single integration. 
About the dark, if it is constant, you can correct your data a posteriori. The same is true if all 4 normalization arrays are constant (discarding dynamic masking). 

But if you want to keep dynamic masking (and you probably need it with your Eiger detector :-) the correction is possible but probably longer than the intergration itself.

pyFAI/ext/CSC_common.pxi
kif: For 1D output, it does not need to fit completely into the cache since one pixel contributes to adjacent bins. This is probably the reason why it is so fast.

2d output will probably not fit into the L1 cache and bins there can be non contiguous...

pyFAI/ext/CSC_common.pxi
kif: If ErrorModel is "no" one can save the histogram of variance for sure and one could save the squared norm as well since it is only used to calculate the std but I also use it internally to check the emptiness of bins. I already encounter a case where the normalization factor (extracted from a beam-stop diode) was negative and crashing ODA pocessing when tested without beam. Since it is one write without (additional) read I prefer to keep it.

So yes, one could save one read and one write.

pyFAI/load_integrators.py
kif: Thanks

Issue #1737, opened 2022-09-27
Refactor geometry engines


kif:
One of the new feature of `pyFAI2` will be that one can plug-in different geometries compared to `pyFAI` which always uses the `poni` geometry and convert the parameters between geometries.

* [x] Make `geometry` a sub-module instead of a module
* [ ] Move Fit2D geometry conversion into different modules
* [ ] Move SPD geometry conversion into different modules
* [ ] Move ImageD11 geometry conversion into different modules
* [ ] ...

PR #1738, opened 2022-09-27
Conversions

kif:
extract geometry conversion from core geometry module and make the available as stand-alone
#1737


kif:
Related to #1737

PR #1815, opened 2023-01-27
Alternative visulalization of geometry

### LINKED ISSUE #1808: [GUI] Improve `pyFAI-calib2`
It would be great in the `Cake&integration` tab, in the right panel which is mostly empty, to display the geometry using the same widget as `pyFAI-integrate`.
2023-01-19-Calibration-7
geometry

kif:
Include a geometry visualization into the integration tab with the ability to change the geometry in pyFAI/FIT2D coordinate system
* saves the modified parameter
* integrated along the modified parameters as well.

close #1808


kif:
Should be refactored on top of the modification introduced by
https://github.com/silx-kit/pyFAI/pull/1821

PR #1821, opened 2023-02-03
Create a widget with just the GeometryTabs

### LINKED ISSUE #1818: [GUI] factorize geometry display/modification between the IntegrationTask in calib2 and GeometryDialog
There is code duplication. One should probably create a common object dealing with this widget and use at the 2 places.
Introduced in #1815 ...

kif:
This is an attempt to mitigate the code duplication introduced in #1815.

The GeometryDialog was re-written to use this "new" widget. Once merged, #1815 can take benefit from this work instead of duplicating the code.

Close #1818


kif:
@vallsv I believe the PR is ready for review. Do you see any feature which would have been broken by this modification ?

PR #1831, opened 2023-02-22
Enable sigma_clip_legacy to be used within a worker

### LINKED ISSUE #1825: Support sigma-clipping with pixel splitting
As far as I understand the current `sigma_clip_ng` implementation does the clipping before the splitting. If you want both it should be the other way around. `sigma_clip` does in the other way around but it is not supported by the `Worker` class. It also gives a deprecation warning.

kif:
close #1825

### DIFF
--- a/pyFAI/azimuthalIntegrator.py
+++ b/pyFAI/azimuthalIntegrator.py
@@ -30,7 +30,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "06/01/2023"
+__date__ = "22/02/2023"
 __status__ = "stable"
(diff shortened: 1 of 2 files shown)


woutdenolf (commented):
I'll check asap. One potential issue that comes to mind is when you want to provide `nbpt_azim` the worker thinks the result is 2D while it is in fact 1D (see `do_2D`).

kif:
Since `ai.sigma_clip_legagy` returns an `Integrate1dResult`, so for the `worker` it is a 1D method, despite it goes via a 2D integration: the worker does not have to be aware of it.

To configure the internal behaviour of `ai.sigma_clip_legagy`  within the worker, on can add the intermediate 2D size in the the `worker.extra_options={"npt_azim": 256, ...}`

kif:
close #1825

woutdenolf:
I tested and it works fine. Thanks @kif !

woutdenolf (approved):
LGTM

PR #1836, opened 2023-03-01
retrieve wavelength only if defined

kif:
related to https://gitlab.esrf.fr/workflow/ewoksapps/ewoksxrpd/-/issues/22

### DIFF
--- a/pyFAI/goniometer.py
+++ b/pyFAI/goniometer.py
@@ -633,7 +633,7 @@ class SingleGeometry(object):
             dict_geo["data"] = self.control_points.getList()
         if self.calibrant is not None:
             dict_geo["calibrant"] = self.calibrant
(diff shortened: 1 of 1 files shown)


woutdenolf:
This works but it exposes another issue: the wavelength is stored in two places the `Calibrant` and in `SingleGeometry` or `GoniometerRefinement` (I suppose in parameterized for). It seems that `SingleGeometry.geometry_refinement.refine3` also updates `Calibrant.wavelength` but this is not the case for `GoniometerRefinement.refine3`.

woutdenolf:
Actually I was wrong: `Calibrant.wavelength` does get updated by `GoniometerRefinement.refine3`.

woutdenolf (approved):
LGTM

Issue #1845, opened 2023-03-13
Most 2D integration do not propagate uncertainties ...


kif:
This varies with Error-models:
* Poisson:
    *  [x]  no histogram python
    *  [X] no histogram cython
    *  [X] no histogram OpenCL
    *  [X] bbox histogram cython
    *  [X] full histogram cython
    *  [x] sparse python (broken)
    *  [x] sparse cython
    *  [x] sparse OpenCL

kif:
I believe this has been fixed

kif:
I quickly checked and all 2D intergator look OK

PR #1847, opened 2023-03-13
Square out 2D integrators:

kif:
* call the proper functions
* validate variance propagation


kif:
For now, they all produce 2D arrays ... most of them are empty.
close #1845

PR #1855, opened 2023-03-28
prevent motor to be null or None

kif:


kif:
make a popup dialog in addition !

PR #1863, opened 2023-04-25
1860 rework massif

### LINKED ISSUE #1860: Changes to massif algorithm from 0.21.3 to 2023.1.0?
Hi,

Dioptas uses the massif algorithm to find peaks for the calibration procedure. Running the test suite with pyFAI version 2023.1.0 and newer shows that massif behaves differently (strictly worse) than before. It finds too many peaks which are off the rings. 

What has changed? Is there a way to get the old behavior back?

Best wishes,
Clemens

kif:
close #1860 

@CPrescher  could you have a look if it fits your needs ?

### DIFF
--- a/pyFAI/detectors/_common.py
+++ b/pyFAI/detectors/_common.py
@@ -33,7 +33,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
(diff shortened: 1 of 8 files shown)


CPrescher:
Hi Jerome,

yes, i think it is good that the default is now false for the extra option. The other changes are related to creating a mask of the gaps, right? I could not test it out, since it does not build completely on windows, but will check it out, when it is released as wheel. Dioptas is now using detectors, but probably not to their full capability. The main reason I added them was for distortion correction.

Best wishes,
Clemens

PR #1877, opened 2023-05-11
Fix upper limit in clamp OpenCL invocations to avoid buffer overflow

alejandrohomsp:
Hi!

By looking at the code, I have the impression that the `clamp`  invocations in the OpenCL code do not account for the correct upper  `DIM - 1` limit of the corresponding array indexes. If this is the case, they do not completely protect against buffer overflows.

What do you think? Is that correct?

Cheers

### DIFF
--- a/pyFAI/resources/openCL/peakfinder.cl
+++ b/pyFAI/resources/openCL/peakfinder.cl
@@ -45,7 +45,7 @@ inline int pf8_calc_buffer2(int dim0, int dim1, int hw){
     int wsy = get_local_size(0);
     int wsx = get_local_size(1);
     int width = wsx + 2 * hw;
-    return (clamp(dim0, -hw, wsy + hw) + hw)*width + clamp(dim1, -hw, wsx + hw) + hw;
+    return (clamp(dim0, -hw, wsy + hw - 1) + hw)*width + clamp(dim1, -hw, wsx + hw - 1) + hw;
 }
 
 
@@ -75,7 +75,7 @@ inline float2 correct_pixel2(
         float radius = radius2d[gid];
         if (isfinite(radius) && (radius>=radius_min) && (radius<radius_max)) {
             float step = radius1d[1] - radius1d[0];
-            float pos = clamp((radius - radius1d[0]) / step, 0.0f, (float)NBINS);
+            float pos = clamp((radius - radius1d[0]) / step, 0.0f, (float)(NBINS - 1));
             int index = convert_int_rtz(pos);
             float delta = pos - index;
             if (index + 1 < NBINS) {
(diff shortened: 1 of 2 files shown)


kif:
Hi Alejandro,

This pattern is occurring twice in a linear interpolator section. I don't think it makes a big difference: If we are on the upper bound, the value is not interpolated (there is a test just after). One could marginally optimize it by calculating the delta only in the "then" section, but on GPU, divergent branches are anyway both executed.

For the other occurance, I have to dive into the code...

alejandrohomsp:
I understand that the event will not happen if the values in the `radius2d` array are inside the `radius1d[0] <= r < radius1d[0] + step * NBINS`. But if due to some bad parametrization `r` goes beyond that limit, then `index = NBINS` and we might get array overflow in when reading `average1d[index]` in the `else` part.

kif:
I had a look at the `pf8_calc_buffer2` function:
This section of the code works on a square patch of 3x3 (hw=1) or 5x5 (hw=2) pixels (typically).
In order to have fast access to neighbouring pixels, a part of the 2D-image is copied in shared memory (1D). 
Mind that 32-1024 pixels are treated simultaneously in adjacent SIMD lanes, so we need to copy only few extra pixels.
This function is used to convert 2d coordinated in the image into the index in the shared buffer.
I believe the "-1" is wrong in this function since indexes are typically -1, 0 and +1 (if hw=1) so removing the upper bound is not wanted. The size of this buffer is `(wg0+2*half_patch)*(wg1+2*half_patch)` . 
Best performances are probably observed with a square patch of 32x32 pixels, and one needs to copy 34x34 pixel in shared memory.

alejandrohomsp:
For `wgX=32` and `hw=1` indices must fit in the `-1 <= i <= 32` range, that is 34 pixels, right? If this is the case, the upper limit is `wgX + hw - 1 = 32`, isn't it? For `hw=2` the range is `-2 <= i <= 33`, with upper limit `wgX + hw - 1 = 33`.

kif:
[Untitled6.pdf](https://github.com/silx-kit/pyFAI/files/11474977/Untitled6.pdf)

I checked the 2D indexation and it looks OK:
* all data got copied
* no bin remains empty
* buffer overflow (would have been spotted in python)
 
so the function calculating the remapping 2D -> 1D is apparently bijective.

alejandrohomsp:
Nice, very elegant validation. Thanks for searching deeper. I am curious about two things:

1. Is the same result obtained if we add `- 1` in both `clamp` expressions?
2. What happen if you change the upper limit of the `range` expressions in the `for` loops with an extra `+ 1`? If it fails, is the same failure seen when changing the `range` lower limit with extra `- 1`?

kif:
Indeed, you are right, the camp does not do its job to prevent the buffer-overflow.

kif:
I think I made a mistake in the last push ...

PR #1898, opened 2023-06-23
Some polish, doc, types

vallsv:
This is a small PR with mostly fixes on doc and types, and small UI polish.

Feel free to tell me what you want to pick, i can create dedicated PR per commits if you prefer.

e78be61ad2d1421ff101900f4348ce5256f1875e is a commit that could in theory introduce a bug.
other commits are pretty safe.

# Tooltips

During the years i have rework my tooltips on Flint.

The patches on `<ul><li>` make the result a bit more compact and readable.

image

# Calibrant selector

The proposal on the calibrant selector make the resulut smaller.

It's a trick to force Qt to use a pure Qt component instead of a widget from the system (or i misunderstand).

image

Instead of 

image

### DIFF
--- a/pyFAI/calibrant.py
+++ b/pyFAI/calibrant.py
@@ -376,7 +376,7 @@ class Calibrant(object):
         """
         h = hash(self._wavelength)
         for d in self.dSpacing:
-                h = h ^ hash(d)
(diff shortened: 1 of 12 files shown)


kif:
LGTM

PR #1902, opened 2023-06-27
Rework pyFAI calibrant as resource

vallsv:
This PR is a proposal to store the calibrant from pyFAI as a resource.

Plus some review of the code, documentation, typing.

Now a calibrant resource from pyfai is stored as `_filename = "pyfai:LaB6"`.
The file path is generated only when it have to be read.

This way it is easy to distinguish calibrant from resources, and other user calibrants.

If one try to change and save such calibrant, it raises an exception.

### DIFF
--- a/pyFAI/calibrant.py
+++ b/pyFAI/calibrant.py
@@ -628,8 +628,9 @@ class Calibrant(object):
             return self._2th.index(angle)
         if delta:
             d2th = abs(numpy.array(self._2th) - angle)
-            if d2th.min() < delta:
-                return d2th.argmin()
+            i = d2th.argmin()
+            if d2th[i] < delta:
+                return i
         return None
 
     def get_max_wavelength(self, index: Optional[int]=None):
--- a/pyFAI/calibrant.py
+++ b/pyFAI/calibrant.py
@@ -34,6 +34,8 @@ Interesting formula:
 http://geoweb3.princeton.edu/research/MineralPhy/xtalgeometry.pdf
 """
 
+from __future__ import annotations
+
 __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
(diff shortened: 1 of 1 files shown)


vallsv:
This break the code if you think people use `.filename` to read files themselfs.

In this case i can propose another patch, to let behave like before.

pyFAI/calibrant.py
```diff
@@ -580,22 +617,24 @@ def get_2th(self):
...
-        """Returns the index in the 2theta angle index
+    def get_2th_index(self, angle: float, delta: Optional[float]=None) -> int:
+        """Returns the index in the 2theta angle index.
 
         :param angle: expected angle in radians
         :param delta: precision on angle
         :return: 0-based index or None
         """
-        if angle and angle in self._2th:
+        if angle in self._2th:
```
kif, line 627: issue are possible if angle is NaN

pyFAI/calibrant.py
```diff
@@ -452,18 +475,19 @@ def _initialize(self):
...
+    def save_dSpacing(self, filename: Optional[str]=None):
         """
-        save the d-spacing to a file
-
+        Save the d-spacing to a file.
         """
         self._initialize()
         if (filename is None) and (self._filename is not None):
+            if self._filename.startswith("pyfai:"):
+                raise ValueError("A calibrant resource from pyFAI can(t be overwritten)")
```
kif, line 490: Typo

pyFAI/calibrant.py
vallsv: I was curious about `bool(NaN)`
```
In [3]: if float("NaN"):
   ...:     print("Hi")
   ...: 
Hi
```

pyFAI/calibrant.py
vallsv: FIxed

kif:
LGTM ... by the way, did you check the notebook on the documentation still run ?
or do they need dome adaptation ?

vallsv:
I haven't checked anything.

The PR is not supposed to break anything, so tell me, if you are fine with the overall idea, i can rework the PR to make it work without regression.

PR #1903, opened 2023-06-27
New calibrant widget

vallsv:
Here is a test to provide a new popup for the calibrant widget.

Related to #1902 

I wanted to speed up the user selection.

First I have tried to provide a minimalistic periodic table of elements to group the calibrants, but it is probably not very convenient to use.

So my guess here is to provide the recently used calibrants.

The code provides:
- A calibrant widget with diffrent selectors
- The last 8 recently used calibrants are displayed
- This also store/restore user calibrants load from files (so no need to search for a file twice)
- A button to load a calibrant from file

image


vallsv:
This PR can be reviewed.

kif:
LGTM

kif:
Thanks for the improvement of this widget.

Issue #1918, opened 2023-08-08
calib2 not compatible with PySide6


malte-storm:
Hello,

while silx allows to use Qt6 (PySide6) as backend, running pyfai-calib2 with PySide6 raises some exceptions because of small differences how signals are handled.

The overloaded signal definition for the list model requires tuples instead of lists in Qt6.

kif:
Thanks a lot ... apparently this addresses the issue #1852
@vallsv  did not find yet time for diving into it and I don't have Qt skills ...
The code LGTM, I am checking if it addresses all issued spotted.

Issue #1852, opened 2023-03-25
pyfai-calib2 does not work with installed pyside6


tjalb:
I'm using a Mac m1 with Python 3.10. In my Python environment I had pyside6 installed and got the following error message when running pyfai-calib2:

```
qt.pysideplugin: Environment variable PYSIDE_DESIGNER_PLUGINS is not set, bailing out.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
Qt WebEngine seems to be initialized from a plugin. Please set Qt::AA_ShareOpenGLContexts using QCoreApplication::setAttribute and QSGRendererInterface::OpenGLRhi using QQuickWindow::setGraphicsApi before constructing QGuiApplication.
qt.pysideplugin: No instance of QPyDesignerCustomWidgetCollection was found.
ERROR:pyFAI-calib2.UNCAUGHT:'xml.etree.ElementTree.Element' object has no attribute 'getchildren'
Traceback (most recent call last):
  File "/opt/homebrew/bin/pyfai-calib2", line 8, in <module>
    sys.exit(main())
  File "/opt/homebrew/lib/python3.10/site-packages/pyFAI/app/calib2.py", line 594, in main
    window = CalibrationWindow(context)
  File "/opt/homebrew/lib/python3.10/site-packages/pyFAI/gui/CalibrationWindow.py", line 152, in __init__
    self.__tasks = self.createTasks()
  File "/opt/homebrew/lib/python3.10/site-packages/pyFAI/gui/CalibrationWindow.py", line 259, in createTasks
    ExperimentTask(),
  File "/opt/homebrew/lib/python3.10/site-packages/pyFAI/gui/tasks/AbstractCalibrationTask.py", line 48, in __init__
    self._initGui()
  File "/opt/homebrew/lib/python3.10/site-packages/pyFAI/gui/tasks/ExperimentTask.py", line 55, in _initGui
    qt.loadUi(pyFAI.utils.get_ui_file("calibration-experiment.ui"), self)
  File "/opt/homebrew/lib/python3.10/site-packages/silx/gui/qt/_pyside_dynamic.py", line 232, in loadUi
    widget = loader.load(uifile)
  File "/opt/homebrew/lib/python3.10/site-packages/silx/gui/qt/_pyside_dynamic.py", line 166, in load
    self._parse_custom_widgets(uifile)
[...]

PR #1919, opened 2023-08-08
Make pyfai-calib2 pyside6 compatible

malte-storm:
Some small changes to make pyfai-calib2 compatible with PySide6.
Fix for issue  #1918

### DIFF
--- a/pyFAI/gui/model/ListModel.py
+++ b/pyFAI/gui/model/ListModel.py
@@ -139,7 +139,10 @@ class ListModel(AbstractModel):
     :param parent: Owner of this model
     """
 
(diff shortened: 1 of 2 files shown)


kif:
LGTM. Fixes the bug
close #1918 
close #1852

vallsv:
Sounds good. Thanks.

Just a question for my knowlesge, is the `tuple` version backward compatible with PyQy5?
I anyway prefer this patch, it's safer.

malte-storm:
Sorry for the late reply, holiday time...

@vallsv: No, unfortunately the tuple is not Qt5 compatible, therefore I added the explicit settings for qt5/6.

Issue #1931, opened 2023-09-05
[Calib2] the detection of the calibrant selection is not working ...


kif:
Apparently the new widget for selecting the calibrant is no more noticing application that the calibrant is selected (and valid).

Spotted on a debian12 build.

kif:
![image](https://github.com/silx-kit/pyFAI/assets/1018880/d6b1f5cd-721b-499e-b1a3-44d33cfb2e68)

kif:
There is also a typo in the contextual help ...

PR #1932, opened 2023-09-05
fix English spelling.

kif:
Related to #1931:
- [X] fix conjugation
- [ ] fix calibrant issue


kif:
I will leave the calibrant signal handling to @vallsv

PR #1948, opened 2023-09-14
numexpr work arround

kif:
I am not especially happy to put in place such work around, I find it pretty ugly but something is needed while waiting for the issue to be fixed upstream.

What I did, theer is a hook to import numexpr from third_party where numexpr is locally patch (or not) to tell it not to sanitize the input. Unfortunately, this has to be done in 2 different ways, using the environment variable for `numexpr.evaluate` and patching the class for `numexpr.NumExpr`. Apparently it works. If you have a better idea, I would be happy to hear it.

### DIFF
--- a/pyFAI/geometry/core.py
+++ b/pyFAI/geometry/core.py
@@ -40,7 +40,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "31/05/2023"
+__date__ = "14/09/2023"
 __status__ = "production"
 __docformat__ = 'restructuredtext'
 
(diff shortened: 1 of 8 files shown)


loichuder:
Since there is only one problematic version, would it be possible instead in the requirements of the project to disallow the version `2.8.6` of `numexpr` ?

Like putting `numexpr != 2.8.6` in `requirements.txt` and other relevant places ?

t20100:
@loichuder proposition is a lot simpler.
Using `numexpr!=2.8.6` in `install_requires` should do the job:

https://github.com/silx-kit/pyFAI/blob/177f7dba3386ef02830f37e9f2588e4d33599efb/setup.py#L982-L991

t20100:
> setup.py is no more used

Indeed, but the same applies for `pyproject.toml`:

https://github.com/silx-kit/pyFAI/blob/177f7dba3386ef02830f37e9f2588e4d33599efb/pyproject.toml#L39C1-L47

kif:
Replaced by solution #1950

PR #1950, opened 2023-09-15
Prevent the usage of the broken numexpr

kif:
Do you prefer this ?
How to ensure this gets propagated to debian/conda/ ... ?


t20100 (approved):

t20100:
> Do you prefer this ?

Yes, much simpler!

> How to ensure this gets propagated to debian/conda/ ... ?

For debian, pyFAI tests should fail so it won't go unnoticed.
For conda, we are maintainers so we can fix it.

Issue #1953, opened 2023-09-21
Pilatus 4


kif:
include spec of this new serie of detectors.

kif:
module size? 513x530 ? 
gap size ? 30x7?
Si or CdTe ? are the gaps different ?

PR #1954, opened 2023-09-21
Implement Pilatus4

kif:


kif:
Waiting for some test images from Dectris to validate the implementation. Qu: what about silicon/CdTe sensors? do they share the same module size? Should we distingate them ?

kif:
close #1953

PR #1956, opened 2023-09-25
Perfrom 2D integration in non-azimuthal space

### LINKED ISSUE #1951: Is it possible to reorder image to rectilinear (qx, qy) coordinates?
Hi all,

I am trying to use pyFAI to do a 2d reordering of detector images, but as target coordinate system I need rectilinear (qx, qy) coordinates instead of polar coordinates (Q, chi). (Such that Q=sqrt(qx^2 + qy^2) and chi=arctan(qy/qx))

I have not found anything like that in the docs and I'm wondering whether it's possible to achieve with the existing methods? Any hints  and pointers in the right direction are greatly appreciated!

Many thanks and cheers!

kif:
Requested to integrate in Qx/Qy space...
close #1951


kif:
This PR deserves many tests ...

kif:
Apparently there is an issue since the azimuthal range is 2pi periodic.

PR #1959, opened 2023-09-29
Take into account periodicity of azimuthal angle

### LINKED ISSUE #1958: 2D integration is limited to 2pi range in vertical dimensions
This is because angles are 2pi-periodic, but other units are not !

kif:
close #1958


kif:
This introduces a regression on the azimuthal range when method.split="bbox" (not "no" or "full") when 2D-integration

PR #1962, opened 2023-10-02
Added convertion from tthrad to RecD2_NM

vallsv:
Here is an update of the helper convertor to handle tth_rad to d^-2.

It's a rework of the info from `pyFAI.units`, but i would be happy to have a double check here.
Could you please check if the math is right?

BTW: I also have that code in Flint for now, to be sure it is working with previous pyFAI versions.

### DIFF
--- a/pyFAI/gui/utils/unitutils.py
+++ b/pyFAI/gui/utils/unitutils.py
@@ -112,5 +112,8 @@ def from2ThRad(twoTheta, unit, wavelength=None, directDist=None, ai=None):
         else:
             beamCentre = ai.getFit2D()["directDist"]  # in mm!!
         return beamCentre * numpy.tan(twoTheta) * 0.001
+    elif unit == units.RecD2_NM:
+        q = (4.e-9 * numpy.pi / wavelength) * numpy.sin(.5 * twoTheta)
+        return (q / (2.0 * numpy.pi)) ** 2
     else:
         raise ValueError("Converting from 2th to unit %s is not supported", unit)


kif:
The math do look OK. Was this feature requested by MX beamlines ?

kif (approved):

kif:
Just checking what's going wrong with RTD. Your PR is OK.

vallsv:
Thanks for the check

> The math do look OK. Was this feature requested by MX beamlines ?

ID15, to display the value in `d^-2` in a tooltip

vallsv:
In fact they prefer to display the stuff in `d*2_A^-2`, there is a registration unit for that, but there is no global named var.

Do you think it is fine if i add the following in the `units` module?

```
RecD2_A = RADIAL_UNITS["d*2_A^-2"]
```

vallsv:
And the equation would be the same as?
```
        q_A = (4.e-10 * numpy.pi / wavelength) * numpy.sin(.5 * twoTheta)
        return (q_A / (2.0 * numpy.pi)) ** 2
```

kif:
The scale factor differs by a factor 100 (i.e. 10²) that's all.

kif:
In this case I would rather do smth like this:
```
elif unit.space == "d*2":
    q_A = (4.e-9 * numpy.pi / wavelength) * numpy.sin(.5 * twoTheta)
    rec_d2_nm =  (q_A / (2.0 * numpy.pi)) ** 2
    return rec_d2_nm/unit.scale
```

kif:
The same could be done to simplify the handling of `q`, `2th` and `r`, each time, just divide by the scale.

PR #1970, opened 2023-10-05
Added d space unit in A

vallsv:
Feedback from ID06.

They expect to use this unit.

I am not quite sure on the description of the formula, but for sure the way to compute it is fine.

```
d_A = 2pi / Q_A-1
```

For now i compute it in a dedicated notebook at ID06, but i will probably have to display it in Flint too, so i think it is better to share it at the same place, if possible.


gudlot:
It is an equation of how to convert q-values to d-spacing.

vallsv:
> It is an equation of how to convert q-values to d-spacing.

Yes, i think it is

vallsv:
Reworked as requested to have ascale 1 with q in meter.

Notice that unit scale for `q` space, and `eq_q()` returns values in `nm^-1` if i am not wrong.

Then i am kind of confuse
- I am not sure if the compensation in the `units` module `d_A.equation=` is right. 
- But the test on `unitutils.py` (which dosn't use `equation`) is for sure fine

kif:
I will add a quick test that validates equation (implemented in `numpy`) and formula implemented in `numexpr` are actually equivalent...

kif:
The extra test in #1971 fails on this unit...
======================================================================
FAIL: test_all (pyFAI.test.test_units.TestUnits.test_all)
----------------------------------------------------------------------
  File "/users/kieffer/workspace-400/pyFAI/build/lib/python3.11/site-packages/pyFAI/test/test_units.py", line 66, in test_all
    self.assertTrue(numpy.allclose(ref,obt), f"Equation and formula do NOT match for {k}")
AssertionError: False is not true : Equation and formula do NOT match for d_A

kif:
I can submit you a patch ...

PR #1983, opened 2023-10-11
Include more recent versions of python in CI

kif:
3.12 for linux
3.11 for windows


kif:
requires silx release

PR #2000, opened 2023-11-20
Correct some errors in geometry calculation + update doc

### LINKED ISSUE #1996: Unit scaling in `center_array` is only applied to the first column/row
Relevant code:

https://github.com/silx-kit/pyFAI/blob/327676400247078e19268426f94fba978004a2bd/src/pyFAI/geometry/core.py#L876

## Short version

Since `center_array`  [expects to work with a 3D array](https://github.com/silx-kit/pyFAI/blob/327676400247078e19268426f94fba978004a2bd/src/pyFAI/geometry/core.py#L886) but I get a 2D array, the 
unit scaling applied to the last dimension of the resulting `center_array` is only applied to the first column/row.

https://github.com/silx-kit/pyFAI/blob/327676400247078e19268426f94fba978004a2bd/src/pyFAI/geometry/core.py#L919

## Long version

Reading `center_array` docstring, the returned array is expected to be 3D.

https://github.com/silx-kit/pyFAI/blob/327676400247078e19268426f94fba978004a2bd/src/pyFAI/geometry/core.py#L886

However, reading the code, the returned `ary` comes from the `unit.equation` which returns an array with the same dimensionality as `x`, `y`, `z`. And `x`, `y`, `z` are 2D arrays (if `shape` is a tuple of 2 integers [as indicated by the docstring](https://github.com/silx-kit/pyFAI/blob/327676400247078e19268426f94fba978004a2bd/src/pyFAI/geometry/core.py#L882)) since `pos` is 3D. So `ary` is in fact 2D...

To add more confusion, the main part of the docstring indicates that the returned array is 2D.

https://github.com/silx-kit/pyFAI/blob/327676400247078e19268426f94fba978004a2bd/src/pyFAI/geometry/core.py#L878

But perhaps am I missing something in my analysis and/or call of `center_array` ?

kif:
close #1996

### DIFF
--- a/src/pyFAI/geometry/core.py
+++ b/src/pyFAI/geometry/core.py
@@ -40,7 +40,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "11/10/2023"
+__date__ = "20/11/2023"
 __status__ = "production"
 __docformat__ = 'restructuredtext'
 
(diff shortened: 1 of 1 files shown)


kif:
Those errors are probably the consequence of copy/paste or replace-all ...

loichuder (approved):
Looks good, thanks !

PR #2013, opened 2023-12-11
Orientation ... still bugs in some calculations

### LINKED ISSUE #2010: `bbox` splitting overestimates the size of the pixels
This bug is a regression from the introduction of `orientation`
Check that the center of pixels falls actually in the middle of the corner, both on `2theta` and `chi`.

kif:
close #2010


kif:
The bug is not addressed

PR #2016, opened 2023-12-13
new arg parser for 1d

EdgarGF93:
- [x] New parser arguments: pixel split, algorithm, implementation...
- [x] Correct typos and prints
- [ ] Update documentation
- [x] Fix integrate2d benchmark

New attributes for "pyFAI-benchmark" tool (only 1d integration works fine so far)

New arguments for argument parser:
- Pixel splitting argument (-ps, --pixelsplit): no, bbox, pseudo, full, all (default=bbox)
- Algorithm argument: histogram (-algo, --algorithm), CSR, CSC, LUT, all (default=histogram, CSR)
- Implementation argument (-i, --impl): python, cython, opencl, all (default=cython, opencl)
- Function argument (-f, --function): legacy, ng, all (default=all)
- all argument (--all) (boolean): if True, benchmarks every available method (default=False)
- Not processor argument (--no-proc): if used, skips the benchmark using the central processor (useful if you want to benchmark only GPU devices using --gpu)
- Some changes in the figure legends to be consistent with the documentation of AzimuthalIntegrator

image

### DIFF
--- a/src/pyFAI/app/benchmark.py
+++ b/src/pyFAI/app/benchmark.py
@@ -67,6 +67,9 @@ def main():
     parser.add_argument("-d", "--debug",
                         action="store_true", dest="debug", default=False,
                         help="switch to verbose/debug mode")
+    parser.add_argument("--no-proc",
+                        action="store_false", dest="processor", default=True,
+                        help="do not benchmark using the central processor")
     parser.add_argument("-c", "--cpu",
                         action="store_true", dest="opencl_cpu", default=False,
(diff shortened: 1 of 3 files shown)


EdgarGF93:
- Tested 2d integration. 
- Clearer plot with legends outside of the axes.
- Vertical dashed lines to indicate the specific detector to benchmark
Figure_1
(after typing `pyFAI-benchmark -f ng -2d -algo csr --gpu`)

PR #2019, opened 2023-12-14
Expose the orientation selection in the calibration GUI

### LINKED ISSUE #2005: [Calib2] Offer the option to chose the orientation of the detector
Not only for the display, but also for the calculation. This option should be well hidden since it is an advanced option, likely to produce incompatible PONI-files

kif:
Close #2005

### DIFF
--- a/src/pyFAI/detectors/orientation.py
+++ b/src/pyFAI/detectors/orientation.py
@@ -38,7 +38,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "04/12/2023"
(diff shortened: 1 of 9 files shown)


kif:
@vallsv  I will need you help to setup a drop down menu for the selection of orientations (from an enum) ... Do you have some time to discuss about the project ?

vallsv:
That's basically that
```
cb = QComboBox()
cb.addItem("Label", userData=YOUR_ENUM_VALUE)
```
But i suggest you to wrap that in a dedicated class, so what the exposed API can only deal with your enum type.

You basically can copy paste and adapt the AutoscaleModeComboBox https://github.com/silx-kit/silx/blob/e1f33d59a25b9edca5ab7c5e13bcd2df2c5a2997/src/silx/gui/dialog/ColormapDialog.py#L232

kif:
Bad push ... sorry -> see #2020

PR #2020, opened 2023-12-18
Expose the orientation selection in the calibration GUI (v2)

### LINKED ISSUE #2005: [Calib2] Offer the option to chose the orientation of the detector

kif:
close #2005


vallsv (approved):

kif:
The modification you suggest has been put into #2027

PR #2063, opened 2024-01-29
Use the dynamic mask in the calibration GUI

kif:
bug reported by Edgar ...


kif:
To be validated manually: it is a GUI stuff
check on an Eiger image to see if all dead pixels are automatically masked

EdgarGF93 (approved):

Issue #2035, opened 2024-01-16
Add grazing-incidence units


EdgarGF93:
Add grazing-incidence units (qxy) and qz (all three dimensions of q and squared root of both horizontal directions squared)

EdgarGF93:
In-plane and Out-of-plane q arrays are now available but with incident_angle and tilt_angle=0.0

- [ ] Method to change manually the angles
- [ ] Integrate along qip and qoop units
- [ ] Join detector orientation with orientation of qip and qoop (instead of using unit_rot90)

kif:
This looks like to be implemented now ... can we close the issue

PR #2066, opened 2024-01-31
added grazing incidence q components

EdgarGF93:
From #2035 
Added the components of vector q in the sample frame to visualize the missing wedge characteristic of grazing incidence experiments. For the moment, incident_angle and tilt_angle are set both to 0.0, but for most Wide-Angle experiments this is enough accuracy to plot the images.
New units:
Qx = "qxgi_nm^-1"
Qy = "qygi_nm^-1"
Qz = "qzgi_nm^-1"
Qxy = "qxgi_nm^-1"
Qxz = "qxgi_nm^-1"
Qyz = "qxgi_nm^-1"

**In pyFAI, z is the beam propagation direction, x/y are the normal/horizontal components.** But upon the rotation(rot3) of the detector, the user will have to choose qxz/qyz  as the in-plane component and qy/qx (respectively) as the surface-normal component.
Check:
https://confluence.esrf.fr/display/~edgar1993a/Introduction+to+Grazing+Incidence+Geometry

 ```
# In-plane component
arr_qyzgi = ai.array_from_unit(unit="qyzgi_nm^-1")
# Out-of-plane component (normal to the sample surface)
arr_qxgi = ai.array_from_unit(unit="qxgi_nm^-1")
plt.scatter(x=arr_qyzgi, y=arr_qxgi, c=DATA, s=0.2, vmin=0, vmax=500)
```
image

### CODE REVIEW

**Comment by kif** [comment]
It would be nice to have the input angles to the equation in radian, alike all other angle (i.e. SI unit). The conversion from deg2rad should be done outside this module.

Is there a way to use integrate2d ?

**Comment by EdgarGF93** [comment]
Fixed an error in the equations. The angles now are in radians. Also, I changed the notation. Now the new units are:
"exitangle_rad" : the vertical scattering angle relative to the incident angle
"qxgi_nm^-1", "qygi_nm^-1", "qzgi_nm^-1"
"qxgirot90_nm^-1", "qygirot90_nm^-1",  "qzgirot90_nm^-1",   (to be used if y positions correspond to the lab horizontal axis)

**But the interesting ones for grazing-incidence are IP (in-plane) and OOP (out-of-plane):**
"qip_nm^-1" (square root of qx and qz)
"qoop_nm^-1" (qy)
"qiprot90_nm^-1" (square root of qy and qz)
"qooprot90_nm^-1" (qx)
So, in principle, you have to take qip and qoop, or qiprot90 and qooporot90

Example:
```
qip = ai.array_from_unit(unit="qiprot90_nm^-1")
qoop = ai.array_from_unit(unit="qooprot90_nm^-1")
plt.scatter(x=qip, y=qoop, c=DATA, s=0.2, vmin=0, vmax=500)
```
image

If we change manually (in the code) to incident_angle=0.18 and tilt_angle=0.2
image

**Comment by EdgarGF93** [comment]
With integrate2d and gui.jupyter.plot2d, the gaps are filled with colours, just as with plt.pcolormesh. plt.scatter is ok but extremely slow:
```
res2d = ai.integrate2d(DATA, 1000,1000, unit=("qiprot90_nm^-1", "qooprot90_nm^-1"))
jupyter.plot2d(res2d)
pass
```
image

kif:
This looks like a joke from the pixel-splitting scheme (bbox by default, the worse possible). Switching off the pixel splitting should help. `plt.scatter` does the histogram in matplotlib, thus the (slow) speed.

EdgarGF93:
Yes, that's it! :)
giwaxs_tilted

Issue #2079, opened 2024-02-13
Wavelength unit incorrect in Geometry.getImageD11


jadball:
The wavelength in ImageD11 parameter files is assumed to be in Angstroms, but pyFAI gives the wavelength in nanometres:

https://github.com/silx-kit/pyFAI/blob/c0773fa13b75c0528a1eee68991e1f3ae92eedd6/src/pyFAI/geometry/core.py#L1556-L1601

EdgarGF93:
Jerome and me are not using ImageD11, so essentially, it would be changing the nm to Angstroms? Is that okay? @jonwright

jonwright:
Yes please.

ImageD11 does not select units but just expects people to be consistent. Most humans choose Angstrom. For ImageD11 you just need to use the same unit for unit cell parameters and wavelength. Also use the same unit for sample to detector distance as pixel size (in the code example above, the distance is micron not nm).

I have in mind there are some other issues with ImageD11 <-> pyFAI parameters. 
- We miss a way to share the spatial distortion information
- there was something missing on the ImageD11 side to receive an ordered dict
- The required flip for the Eiger at ID11 is (-1,0,0,-1) and this is not handled by the pyFAI code above.  
 
@jadball we could probably fix some things on the ImageD11 and then send any needed pull request for pyFAI. I think there is just a sign inversion on some of the tilts for the flip, but flips were recently added on the pyFAI side too. For the distortion it means adding something like dxfile and dyfile entries, or learning to use the nexus detector file on the ImageD11 side. Do you flip the distortion image or assume it is applied before the flip? Yesterday Pierre-Olivier was looking at that story for Dioptas.

For a testcase with ImageD11 (better to skip than add a dependency?). I think both packages can compute pixel positions in 3D space (ImageD11.transform.compute_xyz_lab for pixel centres) and the conventions should be similar:
- ImageD11/fable x axis = beam direction = pyFAI x3 direction?
- ImageD11/fable z axis = vertically up, normal to the beam = pyFAI x1 direction?
- ImageD11/fable y axis = away from the centre of the synchrotron = negative of pyFAI x2 direction?

Comparing two theta is 'easy', for 'azimuth' angle then I guess we have different origins (and maybe signs). Do you have a testcase already? On the ImageD11 side we could try to write a come up with a "to_poni" and/or "from_poni" function too...

kif:
[...]

PR #2080, opened 2024-02-15
Move ImageD11 conversion to an extra sub-module

kif:
This PR tries to address the issues raised in #2079 
- [X] move the ImageD11 conversion code to another submodule 
- [X] Allow the user to select the unit for wavelength and distances (pixel size, sample-detector distance)
- [ ] Use the orientation feature to replace the orientation matrix
- [x] re-wire this new module in pyFAI.geometry (to maintain current API)
- [x] Provide some tests ensuring `ImageD11` and `pyFAI` provide the same geometrical positions


kif:
There is one feature which is not yet implemented (or checked). 
@jonwright maybe we can agree to merge the PR as it is, or closely, and create a new issue + PR to address the change of orientation of the detector ?

jonwright:
Hi Jerome - 
As you can imagine, I spent all my time yak shaving to work out how to look at the branch install and still didn't figure out how to test this against ImageD11 (git). It looks fine to merge by reading it. We just need to add another "feature" to handle the distortion as dxfile+dyfile in here or as a nexus file over in ImageD11. Probably better to fix ImageD11 to read those. 
The "off by 0.5 pixels" is expected, so should be fixable in the test.
I will try to look at the Eiger versus Frelon flips and come up with something for the conversion.
Thanks!
Jon

jonwright (approved):
OK  to merge - I will aim to come back with a pull request if I find a problem, and it is easier for me to figure out how to do git if this is merged!

kif:
There is this new bug for reminding to fix the orientation. Everything else looks OK.

PR #2082, opened 2024-02-21
create RingExtraction class 

### LINKED ISSUE #2078: `extract_cp` loop runs for all calibrant 2 theta values
Hey, I am working on a ring extraction routine at the Australian Synchrotron, and using `SingleGeometry.extract_cp` as a guide. I was trying to speed up the execution time and noticed that the loop for peak extraction in each ring runs many times more than it needed to, because if no `max_rings` is given, it goes over all the dspacings for that given calibrant (https://github.com/silx-kit/pyFAI/blob/main/src/pyFAI/goniometer.py#L681). 
I came out with a simple solution to calculate `tth` as only the calibrant two theta values which are smaller than the maximum value in `geometry_refinement.twoThetaArray`, to decrease unnecessary overhead.

Some other changes I made were:
- I was able to make a `RingExtraction` class from all I gathered from `extract_cp`, breaking it down to smaller tasks and facilitating unit testing;
- I used `ThreadPoolExecutor` from `concurrent.futures` to run ring extraction on multiple rings at a time, reducing execution time from 15s to 6s in my local dev machine.

emilymassahud:
resolves #2078 
This is my implementation of `extract_cp` as a RingExtraction class. The 2 main things I changed here were adding threading inside  `extract_control_points` which runs `extract_list_of_peaks_in_one_ring` concurrently; and `_get_unique_two_theta_values_in_image` which sets `two_theta_values` array to only go to the max two_theta_value encompassed in the calibration image.
As I was making this I did see that @kif created a PR already to speed things up using cython, but I thought I'd send in the PR anyway to see if we could combine them together, as I made a class for it as well. My idea was to make the class to separate the many things `extract_cp` was doing into smaller methods for clarity and making testing easier.

And speaking of tests, I'd like some help running with the unit tests I wrote. We work with pytest as our testing suite, but I couldn't set it up in the project to make them run. I can commit the test module here with all the tests I wrote as well to get help. (just to begin with I am having issues resolving relative imports of my `ring_extraction` module to the test module, and an absolute import throws the error `ModuleNotFoundError: No module named 'pyFAI'`).

### DIFF
--- a/src/pyFAI/ring_extraction.py
+++ b/src/pyFAI/ring_extraction.py
@@ -1,0 +1,349 @@
+"""
+Module used for extracting control points in a calibrant image to be used for geometry refinement.
+"""
+
+import logging
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from typing import Optional
+
+import numpy as np
+from silx.image import marchingsquares
+
+from pyFAI.control_points import ControlPoints
+from pyFAI.goniometer import SingleGeometry
+from pyFAI.massif import Massif
+
+logger = logging.getLogger(__name__)
+
+
+class RingExtraction:
+    """Class to perform extraction of control points from a calibration image."""
+
+    def __init__(
+        self,
+        single_geometry: SingleGeometry,
+        massif: Optional[Massif] = None,
+    ):
+        """
+        Parameters
+        ----------
+        single_geometry : SingleGeometry
+            instance of pyFAI class
+        massif : Optional[Massif]
+            instance of Massif class, which defines an area around a peak; it is used to find
+            neighboring peaks, by default None
+        """
+
+        self.single_geometry = single_geometry
+        self.image = self.single_geometry.image
+        self.detector = self.single_geometry.detector
+        self.calibrant = self.single_geometry.calibrant
+
+        if massif:
+            self.massif = massif
+        else:
+            if self.detector:
+                mask = self.detector.dynamic_mask(self.image)
+            else:
+                mask = None
+            self.massif = Massif(self.image, mask)
+
+        self.two_theta_array = self.single_geometry.geometry_refinement.twoThetaArray()
+        self.two_theta_values = self._get_unique_two_theta_values_in_image()
+
+    def extract_control_points(
+        self,
+        max_number_of_rings: Optional[int] = None,
+        points_per_degree: int = 1,
+    ) -> ControlPoints:
+        """
+        Primary method of RingExtraction class. Runs extract_control_points_in_one_ring for all
+        rings in image, or for max_number_of_rings, if not None.
+
+        Parameters
+        ----------
+        max_number_of_rings : Optional[int]
+            Maximum number of diffraction rings to extract points from, by default None
+        points_per_degree : int, optional
+            number of control points per azimuthal degree (increase for better precision), by
+            default 1
+
+        Returns
+        -------
+        ControlPoints
+            Instance of the pyFAI class with extracted peaks
+        """
+
+        control_points = ControlPoints(calibrant=self.calibrant)
+
+        if max_number_of_rings is None:
+            max_number_of_rings = self.two_theta_values.size
+
+        tasks = {}
+        with ThreadPoolExecutor() as executor:
+            for ring_index in range(
+                min(max_number_of_rings, self.two_theta_values.size)
+            ):
+                future = executor.submit(
+                    self.extract_list_of_peaks_in_one_ring,
+                    ring_index,
+                    points_per_degree,
+                )
+                tasks[future] = ring_index
+
+            for future in as_completed(tasks):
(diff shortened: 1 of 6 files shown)


kif:
* Relative imports are of course preferred over absolute imports.
* The header with the license is needed for the packaging in debian
* Any new submodule/extension should be declared in the `meson.build` so that it gets installed (needed for the testing)

kif:
I believe such a class can be used at 3 different places in pyFAI:
- in pyFAI-calib (the matplotlib calibration tool, still used with notebooks)
- in pyFAI-calib2 (the PyQt5/PySide6 application)
- in SingleGeometry (as demonstrated). 

It is worth investigating an API which fits all of them...

emilymassahud:
Hi Jerome,

Thanks for the instructions, I've done them all. I am adding some tests which I managed to run from the `build` directory calling pytest directly to `test_ring_extraction`. However I am still thinking about how to add it to run in `run_tests.py` because pytest is getting confused with directories (stuff in `build` vs stuff in `src`) and throwing some errors when I call it with `options.test_name`; it throws an `ERROR: file or directory not found: pyFAI.test.test_ring_extraction`, so that'll be a bit of a thinker for me (happy to hear ideas)

The other thing I have to do to `run_test.py` is to discern which modules to run with pytest or unittest (pytest can run all unittest-based tests, but not the other way around). If I just ask it to run all tests, it will run them twice, once with unittest and once with pytest.

emilymassahud:
@kif, after attempting to include pytest to the package, and integrating it to `run_tests.py`, I decided it'd be less work to convert my tests to unittest, so it's all fine now.
However, the ci pipeline is failing because of type hinting of `list`, `dict` and `tuple`, introduced in PEP 585 (>=python3.9), even though pyFAI is supposed to support up to python 3.11 no issues. Can you let me know whether I should use old type hinting syntax, or if it's something we can make work?

I have now to address this to finish it all:

>I believe such a class can be used at 3 different places in pyFAI:
>
>in pyFAI-calib (the matplotlib calibration tool, still used with notebooks)
>in pyFAI-calib2 (the PyQt5/PySide6 application)
>in SingleGeometry (as demonstrated).
>It is worth investigating an API which fits all of them...

kif:
Hi Emily,

Sorry for the delay, I have been off for a week (and not very available this week).

Pytest would need to understand how to build the project ... silx is using pytest so maybe one can get inspiration to adapt from there... but this is another topic. 

PyFAI should now support python 3.12 (silx only supports it since recently) ... and the minimum is currently 3.7. Some type-checking have apparently changed with different . My problem is that ESRF is still using ubuntu 20.04 which uses (still 3.8).
I will check the minimum version for silx  and align pyFAI accordingly.

Thanks for your PR, I'll have a look.

kif:
I fixed some type annotation ... I don't especially like them so I don't master them.

As written previously, such a class is likely to be used at several places:
* in pyFAI-calib2 where the code is available in `pyFAI.gui.helper.RingExtractor.RingExtractorThread`
* in pyFAI-calib where the code is available in `pyFAI.gui.cli_calibration.AbstractCalibration.extract_cpt`

Thus this `RingExtractor` should probably take as input simpler objects:
* 2th positions expected from the calibrant (as a list or numpy array)
* 2th position for each pixel of the image (as a 2d array)
* a mask
* and of course the diffraction image itself
but this deserves some more thinking.

src/pyFAI/ring_extraction.py
```diff
@@ -0,0 +1,388 @@
...
+                mask = None
+            self.massif = Massif(self.image, mask)
+
+        self.two_theta_array = self.single_geometry.geometry_refinement.twoThetaArray()
+        self.two_theta_values = self._get_unique_two_theta_values_in_image()
+
+    def extract_control_points(
+        self,
+        max_number_of_rings: Optional[int] = None,
+        points_per_degree: int = 1,
```
kif, line 96: The number of points per degree is expected to be a floating point number (a value <1 is sometimes desired)

src/pyFAI/ring_extraction.py
```diff
@@ -0,0 +1,388 @@
...
+        """
+        Primary method of RingExtraction class. Runs extract_control_points_in_one_ring for all
+        rings in image, or for max_number_of_rings, if not None.
+
+        Parameters
+        ----------
+        max_number_of_rings : Optional[int]
+            Maximum number of diffraction rings to extract points from, by default None
+        points_per_degree : int, optional
+            number of control points per azimuthal degree (increase for better precision), by
```
kif, line 107: same here: float

kif (commented):
One minor comments: PEP8 insists on the consistency within a project rather than to rules, so
* `numpy` should be fully named (and not called `np`)
* black formatting is making the code far too verbose.

but this is aesthetic and far less important as the code looks good.

PR #2086, opened 2024-03-07
2085_diffmap_nogui

EdgarGF93:
Now, pyFAI-diffmap can be executed without using the GUI and the integration parameters on the worker are updated
Also, modified slightly the function reduce_images in worker and added set_mask method.


kif:
LGTM, thanks

Issue #2094, opened 2024-03-12
Calculation of min_distance_between_control_points in RingExtraction


emilymassahud:
When I created the `RingExtraction `class, I based it on the `SingleGeometry.extract_cp` method, which has a calculation for `min_distance_between_control_points` that I didn't quite understand (it's half the length of `seeds` array, divided by number of points to keep for a given ring). So I implemented a method for the Australian synchrotron which involves the calculation of the length of an arc made by a 1 azimuthal degree in that ring:

![Picture1](https://github.com/silx-kit/pyFAI/assets/81233603/d8734991-e5ec-4bba-9720-9e2707fb30e1)

It just uses the pixel coordinates of a random point in that ring (according to `marching_squares_algorithm.find_pixels`), and the beam centre coordinates. Of course this only works for a circle (which is the configuration of our beamline, no major tilts are provided) or images with small rotations. I thought then to implement something similar for ellipses, but using the minimum distance as the smallest arc we can make with 1 degree for that ellipse, which would be an arc centred around the semi-minor axis of the ellipse.
For that I need help; when pyfai is calculating cartesian positions of peaks based on provided geometry, does calculate the ellipse parameters (semi-minor and major axes, centre and eccentricity) somewhere ? I'll make a branch with my current calculation which expects a circle for clarification.

kif:
Hi Emily,
PyFAI does not calculate the equation of the ellipse since this would assume the layout of pixels in the detector is regular (and most detectors are not -perfectly- regular). 
How can we do then ? 
I use a fairly simple trick on line 707-709:
https://github.com/silx-kit/pyFAI/blob/main/src/pyFAI/goniometer.py#L707

Along each ring, one collects the azimuthal angle of each centre of pixel, in degree & with the precision of an integer. 
Then  keep only one of each numerical value, each roughly represents 1° of arc. So we just count them to measure the length of the arc: simple and efficient. One can easily go to a better precision.

emilymassahud:
I see, thanks for the explanation. But why is `dist_min = len(seeds) / 2.0 / keep` on line 713? https://github.com/silx-kit/pyFAI/blob/main/src/pyFAI/goniometer.py#L713
Is this somehow empirically adjusted? how was this derived?

kif:
I have to admit this is a bit empirical ... I did something quick and it was working pretty well so it remained like this.

[...]

PR #2095, opened 2024-03-12
Calculate min dist between control points in RingExtraction #2094

emilymassahud:
Apart from what I discussed in issue #2094, I also increased max line length of the formatter so the code looks less verbose (as jerome mentioned in a previous comment) and changed from `np` to `numpy`.


kif:
LGTM

PR #2097, opened 2024-03-12
Pilx

kif:
Import project pilx into pyFAI
* [X] Update installer
* [X] Update scripts
* [X] Update headers -> put credit to Loïc
* [X] Call `pilx` as visualization tool for `diffmap`


kif:
The License in the header is (used to be?) needed by Debian for packaging

kif:
The last ting that remains to be done: change the name.
Currently `pyFAI-pilx`  --> `diffmap_view` ?

src/pyFAI/gui/pilx/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.0.4"
```
loichuder, line 1: I think this should be removed ?

loichuder (approved):
All right for me. As discussed, you can change the name `pilx` in whatever you see fit

loichuder:
Also, let's use the latest pilx version to include latest @EdgarGF93 changes: https://gitlab.esrf.fr/silx/pilx/-/releases/v0.0.5

src/pyFAI/gui/diffmap_widget.py
```diff
@@ -568,6 +575,35 @@ def update_processing(self, idx_file, idx_img):
...
+            if self.fig:
+                pyplot.close(self.fig)
+                self.fig = None
+                self.plot = None
+                self.img = None
+                self.axplt = None
+                self.aximg = None
+
+        if start_pilx and isinstance(start_pilx, str) and os.path.exists(start_pilx):
+            self.processingFinished.emit(start_pilx)
```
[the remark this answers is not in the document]
kif: No, because the visualization has to be triggered from the main thread and the processing occurs in a separate thread ... Several options have been considered with Thomas, this is one of them.

src/pyFAI/gui/pilx/__init__.py
kif: Is it used anywhere ?
If not it could be removed.

kif:
The changes made by Edgar have been fixed in another way when I copied the code to pyFAI ... so it does not make much sense.

src/pyFAI/gui/pilx/__init__.py
loichuder: Not it is not used. It can be removed.

src/pyFAI/gui/diffmap_widget.py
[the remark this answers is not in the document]
loichuder: Got it :ok_hand:

loichuder:
> The changes made by Edgar have been fixed in another way when I copied the code to pyFAI

How so ?

kif:
I have seen only 3 commits which comes down to:
* this https://gitlab.esrf.fr/silx/pilx/-/commit/5e59d5d0a7f54efebc66c72bd58386480aa34b6a
* one removal of a space 
* the version number changed ... 

The referenced commit has been replaced by a `pyFAI.load(config)` like in this PR

loichuder:
Ah ok, I didn't realize you made changes to the pilx code compared to 0.0.4.

Btw, I guess you wanted to write `factory` here:
https://github.com/silx-kit/pyFAI/pull/2097/files#diff-b61db08effa9f43c957c7409f8a1281a8c4d190ab1629ef90923f5f573c24ea3R52

PR #2114, opened 2024-03-25
2113 diffmap variables

EdgarGF93:
Now, Diffmap is instantiated in DiffMapWidget in a more general way.
To do that, I had to unify the name of the variables: from "fast_motor_points" to "nbpt_fast", and from "slow_motor_points" to "nbpt_slow"
Also, took the chance to refactor "npt_rad" and "npt_azim" to "nbpt_rad" and "nbpt_azim", respectively, which is the worker nomenclature

### DIFF
--- a/src/pyFAI/app/diff_tomo.py
+++ b/src/pyFAI/app/diff_tomo.py
@@ -68,7 +68,7 @@ class DiffTomo(DiffMap):
         :param nRot: number of translations
         :param nDiff: number of points in diffraction pattern
         """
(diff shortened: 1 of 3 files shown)


kif:
On the one hand, this looks reasonable but on the other it breaks the API, so it is potentially dangerous.

A change of API (used in production) should go via:
* A compatibility layer with the former API
* A deprecation warning of the former API 
* A date of end of support of the former API.

I know that's a lot of code to write, but it is the only way to pay-back some technical dept.

kif:
I made a PR on this topic

PR #2118, opened 2024-04-03
deprecation warning in set_config

EdgarGF93:
Something like this? In set_config, the keys of the incoming dictionary are updated with deprecation warning


kif (approved):
LGTM

PR #2143, opened 2024-04-18
2127 flat, dark, background in pyFAI-calib2 [PROPOSAL]

EdgarGF93:
Proposal for including dark current and flat field images in pyFAI-calib2

- [x] Dark current file(s)
- [x] Flat field file

Idea:
- Currently, pyFAI-calib2 works with ImageModels to handle the data arrays from the file and the mask. **ImageFromFilenameModel** is a generic class that handles the filename and the data array. **MaskedImageModel** computes the masking of the array.
- Idea: to include two new classes:
1. **DarkedMaskedImageModel**, same as MaskedImageModel but also takes an instance of ImageFromFilenameModel for the dark current file, and computes the masking and following subtraction of the dark current array

darked_masked_data = Masking(data_array) - dark_current_array

2. **CorrectedImageModel**, takes two instances of DarkedMaskedImageModel and an instance of ImageFromFilenameModel for the flat field file, and computes the subtraction of both DarkedMaskedImageModel following by a normalization with flat field array. The idea is to use the second instance of DarkedMaskedImageModel as a background file. This is the correction protocol:

corrected_image = (darked_masked_image_data - darked_masked_background_data) / flat_field

inspired by: https://lookingatnothing.com/index.php/archives/2251


src/pyFAI/gui/model/DarkedMaskedImageModel.py
kif, line 1: Can't this be replaced by preprocessedImageModel ?

EdgarGF93: correct, DarkedMaskedImageModel.py is no longer needed

kif:
Is this still WIP ?

EdgarGF93:
Yes, I will test the change of the image and I will label Ready to merge

EdgarGF93:
now, flat and dark images can be uploaded through CLI, dialog file and drag&drop, and they will be properly computed along all the calib2 tasks

kif:
@jonwright are you willing to test it ?

kif (approved):
LGTM

jonwright:
Sure - we will try this later today or during the weekend - sometime soon after we figure out how to install pyFAI from a pull request.

kif:
you can use "./bootstrap pyFAI-calib2" to test locally without installing. 
You may install all dependencies before-hand with a a "pip install -r requirements.txt --upgrade"

PR #2144, opened 2024-04-18
[pilx] multiselection: fix/unfix tool

EdgarGF93:
I changed the add tool to Fix/Unfix curve to the plot. Feedback is welcome ;)

Peek 2024-04-18 20-18


src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -267,44 +290,74 @@ def onMouseClickOnImage(self, x: float, y: float):
             radial_value + self._delta_radial_over_2,
         )
 
-    def getColorLastCurve(self):
-        curves = self._integrated_plot_widget.getAllCurves()
-        if curves:
-            return self._integrated_plot_widget.getAllCurves()[-1].getColor()
+    def getColorCurve(self, legend: str):
```
loichuder, line 293:
```suggestion
    def getCurveColor(self, legend: str):
```
The actual thing you are retrieving is the curve color (or the color of the curve), not a curve.

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -267,44 +290,74 @@ def onMouseClickOnImage(self, x: float, y: float):
...
+        curve = self._integrated_plot_widget.getCurve(legend=legend)
+        if curve:
+            return curve.getColor()
         else:
             return
 
+    def getColorAvailable(self, legend: str):
```
loichuder, line 300:
```suggestion
    def getAvailableColor(self, legend: str):
```

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -267,44 +290,74 @@ def onMouseClickOnImage(self, x: float, y: float):
...
+        if self._integrated_plot_widget.getCurve(legend=f"INTEGRATE{legend}"):
+            color = self._integrated_plot_widget.getCurve(legend=f"INTEGRATE{legend}").getColor()
+        else:
+            index_color = 0
+            color = rgba(self._integrated_plot_widget.getDefaultColors()[index_color])
+            while color in [self._integrated_plot_widget.getCurve(legend=legend).getColor() for legend in self._integrated_plot_widget]:
```
loichuder, line 306: This while loop becomes infinite if there are more than 10 curves pinned.

I think we can instead use the method `_getColorAndStyle` from the `PlotWidget`: https://github.com/silx-kit/silx/blob/755fa6d3cbe35e1078f65557f6e9e8352b826da5/src/silx/gui/plot/PlotWidget.py#L3086

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -267,44 +290,74 @@ def onMouseClickOnImage(self, x: float, y: float):
...
-                    self._integrated_plot_widget.addCurve(
-                        x=radial_background,
-                        y=intensity_curve,
-                        legend=legend,
-                        selectable=False,
-                        color=self._integrated_plot_widget.getCurve(legend=legend).getColor()
-                    )
-                return
+        # Unset the background if it's the same pixel and delete markers
+        if (indices.row, indices.col) in self._intensity_background_curve.keys():
```
loichuder, line 318: Ok I see now why you want a dict. It is to store the indices as well.

In my opinion, you could advantageously use a custom dataclass/namedtuple instead:

```
class Pattern:
   row: int
   col: int
   data: numpy.ndarray
```

And use `None` when there is no background.

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -267,44 +290,74 @@ def onMouseClickOnImage(self, x: float, y: float):
...
+            self._intensity_background_curve = {}
+            markers = [item for item in self._map_plot_widget.getItems() if isinstance(item, MarkerBase)]
+            _ = [self._map_plot_widget.removeMarker(legend=marker.getLegend()) for marker in markers if "BG" in marker.getLegend()]
```
loichuder, line 321: Why not:
```
self._map_plot_widget.removeMarker(legend="BG_LOCATION")
```
?

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -267,44 +290,74 @@ def onMouseClickOnImage(self, x: float, y: float):
...
+        else:
```
loichuder, line 322: Since you don't need to execute the following instructions if the background is unset, you can `return` after removing the corresponding marker.

This `else` and the `if self._intensity_background_curve:` below then become unnecessary

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -267,44 +290,74 @@ def onMouseClickOnImage(self, x: float, y: float):
...
+            if self._intensity_background_curve:
+                intensity_curve_ -= pattern_background
+
+            self._integrated_plot_widget.addCurve(
+                    x=radial_background,
+                    y=intensity_curve_,
+                    legend="INTEGRATE",
+                    selectable=False,
+                    color=self._integrated_plot_widget.getCurve(legend="INTEGRATE").getColor(),
+                    resetzoom=self._integrated_plot_widget.getGraphXLimits() == (0,100),
```
loichuder, line 344: Why is the rationale behind this `resetzoom` condition ?

src/pyFAI/gui/pilx/widgets/IntegratedPatternPlotWidget.py
```diff
@@ -73,6 +73,13 @@ def __init__(self, parent=None, backend=None):
         centralWidget = self._initCentralWidget(self._statusBar)
         self.setCentralWidget(centralWidget)
 
+    def __iter__(self):
+        for legend in self.getAllCurves(just_legend=True):
+            yield legend
+
+    def __len__(self):
+        return len(self.getAllCurves(just_legend=True))
```
loichuder, line 81: Are those needed somewhere ? If not, I would remove them.

src/pyFAI/gui/pilx/MainWindow.py
EdgarGF93: It has to be executed, to plot the curves again without any subtraction

src/pyFAI/gui/pilx/MainWindow.py
EdgarGF93: it's the original shape, so if the user changes it, the zoom will not reset (of course, there is a better way to do this)

EdgarGF93:
I changed several things, now the "Point", i.e. the selection of a pair of coordinates from the map instantiates the class DiffMapViewPoint, and to handle the multi-selection of points, it checks the instances in a list, instead of using a dictionary. Now, the multi-selection and the background subtraction is more 'readable' and easy to maintain

There is still an issue, the method _getColorAndStyle to change the color do not fall into infinite loop because it changes the style of the line but the marker got the same color, so there are duplicated markers in the map.

image

loichuder (commented):
Way better in terms of UX !

Not too sure about the `Point` classes: they seem to be hold a lot of info for not much. I think they should only hold info that is retrieved from the file: the curves and the labels because that the interesting thing to abstract. 

I pushed a commit where I refactored the `Point` class by using `ImageIndices` (which was already there) to hold the fixed and unfixed point coordinates. Tell me what you think.

loichuder:
> There is still an issue, the method _getColorAndStyle to change the color do not fall into infinite loop because it changes the style of the line but the marker got the same color, so there are duplicated markers in the map.

I can live with that for now. If needed, we can rethink the color cycle. I just didn't want it to get in the way of the GUI usage by falling in an infinite loop.

src/pyFAI/gui/pilx/MainWindow.py
loichuder: Doesn't seem to work on my end. I guess it is data-dependant ?

There is probably a programatic way in silx to know if the zoom was changed or not.

loichuder: Like: `len(PlotWidget.getLimitsHistory()) == 0`

EdgarGF93:
Hi, Loic, I thinks it's okay. It is true that we can even simplify more the Point class in the following PR. I also added a clear button on the Map toolbar to remove the fixed points at once. If you agree we can merge this branch and open a new one.

loichuder (approved):
Thanks a lot, this is great :slightly_smiling_face:

kif:
I will merge the PR ... 
when you make comments like "to be fixed in a future PR", please open immediately another issue, labeled "enhancement"  and linked to this PR so that it does not get forgotten.

EdgarGF93:
continue in #2191

Issue #1998, opened 2023-11-20
Create option to apply polarization directly (i.e. not a weighted average)


kif:
For more details, see the discussion with the AzInt team on the pyFAI mailing list. 
- [x] implement in preprocess python
- [x] implement in preprocess cython
- [x] implement in preprocess opencl
- [x] wire in azimuthal integrator as opt-in (i.e. new kwarg option) in 1D
- [x] same in 2D
- [x] non regression test
- [x] documentation (tutorial?)

emilymassahud:
Hi Jerome, I have been in contact with MAX IV scientist trying to understand this, but I don't quite fully grasp this yet, so I'd like to kindly ask a few questions about it..
From `azint` code I saw they apply all the corrections to each individual pixel of the data after polar transformation to radial and azimuthal angle, and then they perform the sparse matrix multiplication using CSR method.

- Now, for pyFAI, is the normalization (which includes polarization, solid angle and flat) applied after binning (radially and, in case of 2D integration, azimuthally)? this is what I think I understood from the CSR engine and the python implementation of `preproc`. Also, correct me if I'm wrong, but if we set `split_result` to False, are the corrections done pixel wise? In case I'm right, then is it necessary to apply polarization correction on raw, if we can just NOT `split_results`?
- Is `preproc` done on the original image, or does one have to apply a polar transformation to the raw image before passing it as arg?
- Even if, in the case of the `CSR_integrator`, which does `split_result=4`, do we expect a significant difference if we apply polarization correction to raw data instead of having it as part of `normalization`? is this difference in intensity, error?
Sorry if any of those sounds silly.. I am just starting to try and understand the algorithm (:
(I hadn't joined the mailing list in November 2023, so I am not fully aware of what was discussed then)

kif:
Contribution from the MaxIV diffraction & scattering group
[errorPropAziInt.pdf](https://github.com/silx-kit/pyFAI/files/14934960/errorPropAziInt.pdf)

kif:
This is all done now. See the documentation in 
https://github.com/silx-kit/pyFAI/tree/main/doc/source/usage/tutorial/Variance/Unweighted.ipynb

PR #2148, opened 2024-04-25
Make most of the attributes of `method` private and unmutable and introduce `weighted_average`

kif:
Unlike the other attributes, this one is mutable since any algo can manage the two. The difference is made at the pre-processing stage of the integration.


kif:
Related to other #1998

PR #2193, opened 2024-06-07
Provide two distinct origins when the detector is flipped

### LINKED ISSUE #2177: [calib2] Distinguish the origin of the detector from the pixel indexed (0,0)

kif:
One for the image and another for the detector.
close #2177


kif:
Example of image plot:
image

Example of integrated plot:
image

kif:
I am not completely happy, having X and Y vectors would have been better...

PR #2200, opened 2024-06-13
[pilx] new features

EdgarGF93:
- [x] Plot with 10 available colors. Before, there was the period was 5 colors, because even if you provide the addCurve method with a color, if the curve is new, the silx method request a color again, so every time we plot a new curve, the index of the color advances 2 steps. So, at the time of adding a curve, I let the addCurve request the next color, then we retrieve manually that color and use it to plot the marker
- [x] Viridis as default colormap
- [x] Mask is applied on the pattern, only if pyFAI-diffmap saves the dataset 'maskfile', which is not happening on every situation (open another issue)
- [x] Re-wire pyFAI-diffmap to incorporate masks in the output file

### DIFF
--- a/src/pyFAI/diffmap.py
+++ b/src/pyFAI/diffmap.py
@@ -293,6 +293,8 @@ If the number of files is too large, use double quotes like "*.edf" """
             mask = urlparse(options.mask).path
         elif config.get("do_mask", False) or config.get("mask_file", None):
             mask = urlparse(config.get("mask_file", None)).path
+        else:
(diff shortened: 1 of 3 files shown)


EdgarGF93:
image

kif (approved):
LGTM
Nevertheless  entries looks hard coded to `entry_0000` and images to `images_0001` which is not certain. most beamline use several input HDF5 files and the program should be able to cope with several entries per file.

PR #2207, opened 2024-06-14
Add "absorption" kwarg to integrate1d_ng and sigmaclip_ng

pierrepaleo:
This PR makes the `absorption` parameter usable from high-level functions `integrate1d_ng` and `sigmaclip_ng`. 

For `integrate1d_ng` there are lots of different class instantiations depending on the model/implementation/etc ; not sure all of them support the `absorption` parameter.

Also perhaps this parameter should be made visible to other functions.


kif:
I believe it would make sense to have this *absorption* parameter just after *dark*, *flat* ...

kif (changes requested):
Could you please group *absorption* with flat/polarization ...

Note PR #2209 wires absorption in the engines which were missing it.

PR #2210, opened 2024-06-17
export orientation also for nexus detectors

kif:
config should contain the kwargs of the constructor


kif:
Generates #2213

mjdiff (commented):
When f"PixelSize= {self._pixel1:.3e}, {self._pixel2:.3e} m" we get for 50 um pixel 
5.000e-05 m

more "natural"  would be  50.00e-6 m  as we always work using um

PR #2215, opened 2024-06-20
pilx: Display frame index as plot title

loichuder:
image

### DIFF
--- a/src/pyFAI/diffmap.py
+++ b/src/pyFAI/diffmap.py
@@ -427,6 +427,7 @@ If the number of files is too large, use double quotes like "*.edf" """
         process_grp["dim1"].attrs["range"] = self.slow_motor_range
         process_grp["dim2"] = self.nbpt_rad
         process_grp["dim2"].attrs["axis"] = "diffraction"
+        process_grp["offset"] = self.offset
         config = nxs.new_class(process_grp, "configuration", "NXnote")
         config["type"] = "text/json"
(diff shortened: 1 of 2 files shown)


EdgarGF93:
Do we need to take into account the offset of pyFAI-diffmap?
I mean, if there is an offset of 5 points, do they want to see that Frame as 0 or 5?

EdgarGF93:
And I think the correct equation is:
```
image_index = row * map_shape[1] + col
```

If I integrate with pyFAI-diffmap -t 12 -r 3 (12 points fast motor, 3 points slow motor). The last point of the first row (fast-motor) should be 11, 
image

loichuder:
> And I think the correct equation is:
> 
> ```
> image_index = row * map_shape[1] + col
> ```
> 
> since max_shape = [points_slow_motor, points_fast_motor, radial_bins]

Probably. You know this better than me.

> Do we need to take into account the offset of pyFAI-diffmap?

Depends. The frame index is used to index the `/entry_0000/measurement/images_0001` dataset. Is it saved with or without offset ?

kif:
I confirm that:
`image_index = row * map_shape[1] + col + offset`
How is it managed when several input files are provided ?

loichuder:
> How is it managed when several input files are provided ?

It is not managed. I haven't encountered that usecase yet.

EdgarGF93 (approved):

Issue #2227, opened 2024-07-01
Ewoks does not execute integration with generic Detector


EdgarGF93:
Testing at ID13, pyFAI integration workflow with poni file using generic detector.
The poni file was generated with pyFAI-calib2, so the shape is correct. 
I would bet the problem is in the worker (ewoks integrate uses pyFAI worker)

```
# Nota: C-Order, 1 refers to the Y axis, 2 to the X axis
# Calibration done at Mon Jul  1 16:57:28 2024
poni_version: 2.1
Detector: Detector
Detector_config: {"pixel1": 9.999999999999996e-05, "pixel2": 9.999999999999996e-05, "max_shape": [576, 748], "orientation": 3}
Distance: 0.0
Poni1: -0.9999999999999997
Poni2: -0.4534043661362292
Rot1: -0.7203892792219219
Rot2: 0.9602593871292956
Rot3: 0.0

```

```
Traceback (most recent call last):
  File "/home/opid13/.conda/envs/ewoksworker/lib/python3.9/site-packages/ewokscore/task.py", line 401, in execute
    self.run()
  File "/home/opid13/local/ewoks/ewoksxrpd/src/ewoksxrpd/tasks/integrate.py", line 411, in run
    result = worker.process(
  File "/home/opid13/.conda/envs/ewoksworker/lib/python3.9/site-packages/pyFAI/worker.py", line 403, in process
    raise err
  File "/home/opid13/.conda/envs/ewoksworker/lib/python3.9/site-packages/pyFAI/worker.py", line 378, in process
    integrated_result = self._processor(**kwarg)
  File "/home/opid13/.conda/envs/ewoksworker/lib/python3.9/site-packages/pyFAI/azimuthalIntegrator.py", line 1161, in integrate1d_ng
    solidangle = self.solidAngleArray(shape, correctSolidAngle)
  File "/home/opid13/.conda/envs/ewoksworker/lib/python3.9/site-packages/pyFAI/geometry/core.py", line 1257, in solidAngleArray
    dssa = numpy.fromfunction(self.diffSolidAngle,
  File "/home/opid13/.conda/envs/ewoksworker/lib/python3.9/site-packages/numpy/core/numeric.py", line 1845, in fromfunction
    return function(*args, **kwargs)
  File "/home/opid13/.conda/envs/ewoksworker/lib/python3.9/site-packages/pyFAI/geometry/core.py", line 1229, in diffSolidAngle
[...]

PR #2231, opened 2024-07-04
Generic detector in worker

kif:
related to #2227

### DIFF
--- a/src/pyFAI/io/integration_config.py
+++ b/src/pyFAI/io/integration_config.py
@@ -31,7 +31,7 @@
 __author__ = "Jerome Kieffer"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "15/04/2024"
+__date__ = "04/07/2024"
 __docformat__ = 'restructuredtext'
 
 import logging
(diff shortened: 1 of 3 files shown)


EdgarGF93 (commented):
LGTM, thanks, I agree that is not ideal.
The thing is, is there a way to make a difference between a configuration dictionary that is version '1' and one manually defined dictionary with correct keys (like in this case) ?

So for example, with 
```
In [7]: integration_options_generic = {
   ...:        'detector': 'Detector',
   ...:        'detector_config': {'pixel1': 9.999999999999999e-05,
   ...:         'pixel2': 9.999999999999999e-05,
   ...:         'max_shape': [576, 748],
   ...:         'orientation': 3},
   ...:        }

In [8]:

In [8]: worker = Worker()

In [9]: worker.set_config(integration_options_generic)

In [11]: worker.ai.detector
Out[11]: Undefined detector

```
So, this way, this dictionary would not fall into _patch_v1_to_v2 method, and would be properly normalized (because the keys are correct)

kif:
Your dict defines a poni/geometry, not a worker. 
It is  mistake to try to ingest an (incomplete)  poni to a worker.

EdgarGF93 (approved):

PR #2242, opened 2024-07-22
[2233] worker_version_4

EdgarGF93:


src/pyFAI/gui/widgets/WorkerConfigurator.py
```diff
@@ -550,15 +578,9 @@ def __savePoni(self):
...
         filename = dialog.selectedFiles()[0]
-        config = self.getConfig()
-        if config.get("wavelength") is None:
-            config.pop("wavelength")
-        from ...detectors import detector_factory
-        detector = detector_factory(config.get("detector"), config.get("detector_config"))
-        from ...geometry import Geometry
-        ai = Geometry(detector=detector)
-        ai.set_config(config)
-        ai.save(filename)
```
EdgarGF93, line 561: Here, I'm not sure why saving the .poni file has to include all these lines (to be discussed)

PR #2252, opened 2024-08-26
get_config_from_poni

EdgarGF93:


EdgarGF93:
yes, we can add for the moment just a simple template in the PoniFile class

kif (approved):
LGTM

PR #2254, opened 2024-08-26
Generic CrystFEL geometry file parser

### LINKED ISSUE #2253: Import geometry from crystfel format
Opposite  of #1816

kif:
misses some regression tests ...

Close #2253

### DIFF
--- a/src/pyFAI/detectors/_common.py
+++ b/src/pyFAI/detectors/_common.py
@@ -33,7 +33,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "25/06/2024"
+__date__ = "27/08/2024"
 __status__ = "stable"
 
 import logging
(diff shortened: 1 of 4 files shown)


kif:
* [ ] tests: Ensure all geometry files from https://gitlab.desy.de/thomas.white/crystfel/-/tree/master/doc/examples are properly parsed.
* [ ] Ensure all pixels not in any module are masked out.

kif:
# Important note about coordinate conventions

Note that CrystFEL considers all peak locations to be distances from the corner
of the detector panel, in pixel units, consistent with its description of
detector geometry (see the section about corner_x above).  The software
which generates the HDF5 or CXI files, including Cheetah, may instead consider
the peak locations to be pixel indices in the data array.  In the former case,
a peak position (0,0) corresponds to the very corner of the detector panel.  In
the latter case, position (0,0) corresponds to the center of the first pixel,
and the very corner would be (-0.5,-0.5).
To compensate for this discrepancy, CrystFEL will, by default, add 0.5 to all
peak coordinates. See the indexamajig option --no-half-pixel-shift if
this isn't what you want.

mjdiff (approved):
line 68 : just pass after except can be misleading

PR #2260, opened 2024-09-02
test without compiling pyopencl

kif:


kif:
@t20100 Like this it is working with the intel driver ... there was a library missing in the package. All test of pyFAI pass successfully (i.e. not skipped). 
The trick is to `bash intel_installer` then to `export OCL_ICD_VENDORS=$(pwd)/....` rather than to `source  intel_installer`.

PR #2266, opened 2024-09-06
Compatibility between GI and MultiGeometry

EdgarGF93:
- [x] MultiGeometry accepts tuple of units for integrate2d (looking at grazing-incidence)

- [x] If None, radial and azimuthal ranges are calculated and set automatically

- [x] Compatibility with gui.jupyter.plot2d

### DIFF
--- a/src/pyFAI/multi_geometry.py
+++ b/src/pyFAI/multi_geometry.py
@@ -169,8 +169,10 @@ class MultiGeometry(object):
         normalization = numpy.zeros_like(signal)
         count = numpy.zeros_like(signal)
         variance = None
-        self.radial_range = self.radial_range or self._guess_radial_range(lst_data=lst_data)
-        self.azimuth_range = self.azimuth_range or self._guess_azimuth_range(lst_data=lst_data)
+        if self.radial_range is None:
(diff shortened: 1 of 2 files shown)


EdgarGF93:
I copied some lines from Azimuthalntegrator.integrate2d about how to deal with tuple of units. Define a tuple of units all the time, by default, the azimuthal unit is TTH_DEG. If `mg.integrate1d` use the self.radial_unit, if `mg.integrate2d`, use self.unit.

mjdiff (approved):

kif:
I am curious if all notebooks of the tutorial would be running fine with this modification

EdgarGF93:
good point, it unveiled a mistake

kif:
Edgar, to answer your question, there is no simple way to guess the radial/azimuthal range.
One could have a method that measures it, taking the bounding values for all integrators, but it is likely to take a bit of time since it needs to initialize the geometry for all AzimuthalIntegrators.

Since `MultiGeometry` is getting popular, maybe it is a good idea to create such a method and use it if the user does not provide anything sensible.

kif (approved):
LGTM

PR #2268, opened 2024-09-10
Discard peaks with masked pixels

### LINKED ISSUE #2267: peakfinder, discard probable false positive.
Discard any peak if there are masked pixels in the local neighbourhood.

Affects mostly Jungfrau detector where masked pixel have value 0

kif:
Close #2267


kif:
validated on experimental data

PR #2276, opened 2024-09-17
Allow multiple sources for the dataset

### LINKED ISSUE #2275: [pilx] reads only the first HDF5 source dataset
Only supports source data coming from a single HDF5 dataset. 
Typical maps come from several HDF5 files
Using a virtual-dataset would be an option but HDF5 then keeps all files open so it is likely to crash because too many files opened. To me it is not an option when working with a large number of files like we do in this kind of experiments.
I would prefer to use external links so that we can manage the file opening/closing. 

@loichuder I don't know if you have time to look after this issue or if you prefer Edgar or I do it.

kif:
close #2275

### DIFF
--- a/src/pyFAI/gui/pilx/MainWindow.py
+++ b/src/pyFAI/gui/pilx/MainWindow.py
@@ -156,7 +156,7 @@ class MainWindow(qt.QMainWindow):
             try:
                 image_grp = h5file[path]
             except KeyError:
-                error_msg = f"Cannot access diffraction images at {path}: not such path."
+                error_msg = f"Cannot access diffraction images at {path}: no such path."
                 logging.warning(error_msg)
                 status_bar = self.statusBar()
                 if status_bar:
--- a/src/pyFAI/gui/pilx/MainWindow.py
+++ b/src/pyFAI/gui/pilx/MainWindow.py
@@ -152,7 +152,7 @@ class MainWindow(qt.QMainWindow):
             else:
                 self._offset = 0
 
-            # treat source dataset ...
+            # Find source dataset paths
             try:
                 image_grp = h5file[path]
             except KeyError:
--- a/src/pyFAI/gui/pilx/MainWindow.py
+++ b/src/pyFAI/gui/pilx/MainWindow.py
@@ -33,7 +33,7 @@ __authors__ = ["Loïc Huder", "E. Gutierrez-Fernandez"]
 __contact__ = "loic.huder@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "18/04/2024"
+__date__ = "17/09/2024"
 __status__ = "development"
 
 
@@ -41,6 +41,7 @@ from typing import Tuple
 import h5py
 import logging
 import os.path
+import posixpath
 from silx.gui import qt
 from silx.gui.colors import Colormap
 from silx.image.marchingsquares import find_contours
@@ -118,7 +119,17 @@ class MainWindow(qt.QMainWindow):
                  ):
 
         self._file_name = os.path.abspath(file_name)
-        self._dataset_path = dataset_path
+
+        c=0
+        for i in dataset_path[-1::-1]:
+            if i.isdigit():
+                c+=1
+            else:
+                break
+
+        _dataset_path = dataset_path[:-c]
+        path, base = posixpath.split(_dataset_path)
+        self._dataset_paths = {}
         self._nxprocess_path = nxprocess_path
 
         self.sigFileChanged.emit(self._file_name)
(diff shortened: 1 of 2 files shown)


src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -118,7 +119,17 @@ def initData(self,
...
+        c=0
+        for i in dataset_path[-1::-1]:
+            if i.isdigit():
+                c+=1
+            else:
+                break
+
+        _dataset_path = dataset_path[:-c]
+        path, base = posixpath.split(_dataset_path)
+        self._dataset_paths = {}
```
loichuder, line 132: This could go with the modifs l.155 to keep the relevant code together.

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -118,7 +119,17 @@ def initData(self,
...
 
         self._file_name = os.path.abspath(file_name)
-        self._dataset_path = dataset_path
+
+        c=0
+        for i in dataset_path[-1::-1]:
+            if i.isdigit():
+                c+=1
+            else:
+                break
```
loichuder, line 128: I guess we can assume more on the dataset path? 

Do something like `i = dataset_path.find("images_"); _dataset_path = dataset_path[i+6:]; `

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -141,6 +152,31 @@ def initData(self,
             else:
                 self._offset = 0
 
+            # treat source dataset ...
```
loichuder, line 155:
```suggestion
            # Find source dataset paths
```

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -141,6 +152,31 @@ def initData(self,
...
+            try:
+                image_grp = h5file[path]
+            except KeyError:
+                error_msg = f"Cannot access diffraction images at {path}: not such path."
```
loichuder, line 159:
```suggestion
                error_msg = f"Cannot access diffraction images at {path}: no such path."
```

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -141,6 +152,31 @@ def initData(self,
...
+                logging.warning(error_msg)
+                status_bar = self.statusBar()
+                if status_bar:
+                    status_bar.showMessage(error_msg)
+            else:
```
loichuder, line 164: It is really worth continuing the execution of the function if the image group cannot be accessed ?

loichuder (changes requested):

src/pyFAI/gui/pilx/MainWindow.py
kif: I would prefer assume as little as reasonably possible. Why enforce the basename if not needed ?

src/pyFAI/gui/pilx/MainWindow.py
kif: Can make sense ... my idea was to keep the context manager part as small as possible, i.e. put this outside the context manager. No problem for moving it into the `with hdf5file` section.

src/pyFAI/gui/pilx/MainWindow.py
kif: There are plenty of `self.blabla` initialized later on. Putting a return there would prevent them from being initialized. I have no clue what would be the behaviour of the application. You are the author more than me. 
If you believe it is safe to return there, I see no objection.

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -141,6 +142,41 @@ def initData(self,
             else:
                 self._offset = 0
 
+            # Find source dataset paths
+            cnt = 0
+            for char in dataset_path[-1::-1]:
```
loichuder, line 147: Not sure the first `-1` is needed (no need to remove the first character)

src/pyFAI/gui/pilx/MainWindow.py
loichuder: Let's leave it like that, I'll check later.

loichuder (approved):

src/pyFAI/gui/pilx/MainWindow.py
EdgarGF93: Sometimes when I was writing and testing the widget, I moved the file or deleted the raw data, so there was no 2D pattern displayed but I still could test the map and the integrated pattern. I think the program should continue the execution, at least if we cannot see the data, we know immediately that there is a problem with the path of the data

src/pyFAI/gui/pilx/MainWindow.py
loichuder: I mean, we are still enforcing a particular format of basename here (it must have digits at the end and be contained in a group). So I thought that putting stronger conditions would allow the parsing to be more explicit.

src/pyFAI/gui/pilx/MainWindow.py
kif: Actually it is the same, at least since python 2.5. I did not update my knowledge base for a while.

src/pyFAI/gui/pilx/MainWindow.py
EdgarGF93: it would suppose changing the initData attributes again, but if we pass `group_path` and `dataset_name` on the initData method, with `dataset_name` None as default, and in this case, just iterates a get every dataset inside the group. That would be a general solution

src/pyFAI/gui/pilx/MainWindow.py
loichuder: Indeed :+1:

src/pyFAI/gui/pilx/MainWindow.py
kif: Loic, it just imposes all datasets have the same prefix. If not, the `sort`operation will not work as expected.
It is really a pity there is no way to enforce the order of dataset/groups in an HDF5 file. Those are just work around for this weakness. 

Edgar, I am not in favour of changing the signature of methods which are supposed to be in the public API, even though I have no clue if anybody else is using it. As Linus said in https://lkml.org/lkml/2012/12/23/75 ... don't break users' programs. I like Linus because I feel really moderated in comparison with him :)

PR #2282, opened 2024-09-19
integrate2d_fiber API

EdgarGF93:
Straightforward methods to get the qip-qoop map, without explicit method without pixel splitting and without instantiating the units (same as integrate_grazing_incidence or integrate_fiber) but in 2D
Example:
```
res2d_gi = ai.integrate2d_grazing_incidence(data=data, npt_horizontal=1000, incident_angle=0.12*3.14/180, tilt_angle=0.0, sample_orientation=2)
```
image

### DIFF
--- a/src/pyFAI/azimuthalIntegrator.py
+++ b/src/pyFAI/azimuthalIntegrator.py
@@ -1879,7 +1879,7 @@ class AzimuthalIntegrator(Geometry):
 
     def integrate2d_fiber(self, data,
                           npt_horizontal, horizontal_unit=units.Q_IP, horizontal_unit_range=None,
-                          npt_vertical=100, vertical_unit=units.Q_OOP, vertical_unit_range=None,
+                          npt_vertical=1000, vertical_unit=units.Q_OOP, vertical_unit_range=None,
                           sample_orientation=None,
                           filename=None,
                           correctSolidAngle=True,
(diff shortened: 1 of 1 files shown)


src/pyFAI/azimuthalIntegrator.py
```diff
@@ -1877,6 +1877,158 @@ def integrate_grazing_incidence(self, data,
...
+        if isinstance(horizontal_unit, units.UnitFiber):
+            if horizontal_unit.sample_orientation != sample_orientation:
+                horizontal_unit.set_sample_orientation(sample_orientation)
+                reset = True
+        else:
+            horizontal_unit = units.to_unit(horizontal_unit)
+            horizontal_unit.set_sample_orientation(sample_orientation)
+            reset = True
+
+        if reset:
```
kif, line 1914: Please register the reason for the reset and emit a warning about why it is reset:
It can be very difficult to debug slow performances if there is unattended reset of the azimuthal integrator.

src/pyFAI/azimuthalIntegrator.py
```diff
@@ -1877,6 +1877,158 @@ def integrate_grazing_incidence(self, data,
...
+            self.reset()
+
```
kif, line 2015: Same here

kif (approved):
LGTM

Issue #2283, opened 2024-09-19
Move fiber diffraction to another module


kif:
Fiber/grazing incidence is already 200 loc (line of code) in the `AzimuthalIntegrator` class and it is not really doing azimuthal integration. The PR https://github.com/silx-kit/pyFAI/pull/2282 will add 150 loc in addition.

AzimuthalIntegrator suffers chronically from the ["god-object" antipattern](https://en.wikipedia.org/wiki/God_object). In the past, it has already been cut twice to separate the Geometry (using inheritance) and the Detector (using composition).

What about putting FiberIntegrator into another class inheriting from AzimuthalIntegrator, into another module ?
This will require the modification of all the documentation on the one hand and to warn the (few) users on the other. Since the feature is fairly new but likely to be appreciated, this has to be ASAP to avoid breaking users' code.

kif:
One needs to find a "clean" way to promote an instance, for example of "Geometry" into "AzimuthalIntegrator" or "FiberIntegrator" ... maybe with a promote method.

kif:
Idea: create an `integrators` sub-module with `azimuthal` and `surface` modules holding `AzimuthalIntegrator` and `SurfaceIntegrator`.
The pyFAI.azimuthalInetgrator modules becomes completely deprecated.

kif:
## Promotion
add a `geo.promote(type)` method could do a deep-copy of the `Geometry` (or `PoniFile`) instance in to an `Integrator` object (which inherits from `Geometry`)

Care should be taken about `GeometryRefinement` which inherits from `AzimuthalIntegrator`. The refactoring of this is part of pyFAI2 project.

PR #2299, opened 2024-10-08
Method to promote a geometry instance into an azimuthal integrator

kif:
Related to #2283


kif:
@t20100 could you provide me your point of view on this ? 

Edgar will need to promote one class instance into one of its sub-class ...

PR #2308, opened 2024-10-18
Grazingincidence good units

EdgarGF93:


kif:
@EdgarGF93, is this PR ready to be reviewed ?
You should then remove the tag WIP and replace it with Ready ...

EdgarGF93:
Not yet, I want to write a test with the new integrator

kif:
:-)

src/pyFAI/integrator/common.py
```diff
@@ -137,6 +137,7 @@ def __init__(self, dist=1, poni1=0, poni2=0,
 
         self._lock = threading.Semaphore()
         self.engines = {}  # key: name of the engine,
+        self._cache_parameters = {}
```
EdgarGF93, line 140: is it ok to this at this level?
it helps me with caching the grazing angles

src/pyFAI/integrator/fiber.py
```diff
@@ -26,44 +26,131 @@
...
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "09/10/2024"
+__date__ = "24/10/2024"
 __status__ = "stable"
 __docformat__ = 'restructuredtext'
 
 import logging
 logger = logging.getLogger(__name__)
 import numpy
+import matplotlib.pyplot as plt
```
kif, line 40: It is not a good idea to import `matplotlib` or `gui` in a computationnal module

src/pyFAI/integrator/fiber.py
```diff
@@ -156,33 +243,71 @@ def integrate_fiber(self, data,
         result._set_method = res.method
         result._set_compute_engine = res.compute_engine
 
+        if plot_integrated_area:
```
kif, line 246: What about the SoC ? this is a module doing processing, here you create a dependency on plotting. Users won't be happy because it creates a dependency on matplotlib hence qt ...

src/pyFAI/integrator/fiber.py
```diff
@@ -196,269 +321,196 @@ def integrate_grazing_incidence(self, data,
...
-                output_unit.set_tilt_angle(tilt_angle)
-                logger.info(f"Tilt angle set to {tilt_angle} for unit {output_unit}. AzimuthalIntegrator will be reset.")
-                reset = True
-        else:
-            output_unit = units.to_unit(output_unit)
-            output_unit.set_tilt_angle(tilt_angle)
-            logger.info(f"Tilt angle set to {tilt_angle} for unit {output_unit}. AzimuthalIntegrator will be reset.")
-            reset = True
+        if "npt_output" or "npt_integrated" in kwargs:
+            logger.warning(f"npt_output and npt_integrated are deprecated parameters. Use npt_oop and npt_ip instead")
```
kif, line 331: 2 things:
* Create a compatibiliy layer for now
* declare since when this is deprecated and when it will be removed (~2 years).
Have a look at the `pyFAI/utils/decorator` module which contains some helper for deprecation warning.

src/pyFAI/test/test_fiber_integrator.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+class TestFiberIntegrator(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        dist = 0.1
+        poni1 = 0.02
+        poni2 = 0.02
+        detector = detector_factory("Eiger2_4M")
```
kif, line 55: The time spent for running test is large, keep the amount of work as small as (reasonably) possible. Is it possible to reduce the size of the detector to "pilatus100k" for example ? Each test would be 40x faster.

src/pyFAI/integrator/common.py
EdgarGF93: with an __init__ method ?

kif: Yes, just call the parent's constructor and it will work.

kif (approved):
LGTM

PR #2309, opened 2024-10-23
fix SyntaxWarning

kif:
Bug 1085805 in debian


kif:
For info, this is only about formatting doc-strings.

PR #2324, opened 2024-11-05
lighter FiberIntegrator class

EdgarGF93:
Since the fiber unit contains parameters (incident_angle, tilt_angle and sample_orientation), there has to be more than one object


EdgarGF93:
now, the tests should pass but I want to test it more

kif (approved):
LGTM

Issue #2304, opened 2024-10-17
Compatibility between `integrator.fiber` and `multigeometry`


EdgarGF93:
- [ ] Make compatible the implicit instantiation of fiber integrators inside a `Multigeometry` instance
- [ ] Make test

EdgarGF93:
Closed with 2025.01

PR #2331, opened 2024-11-12
Compatibility between FiberIntegrator and MultiGeometry

EdgarGF93:
From #2304 
Wrapping the fiber methods inside the MultiGeometry class


src/pyFAI/multi_geometry.py
```diff
@@ -342,3 +343,493 @@ def reset(self, collect_garbage=True):
...
+                * None for no correction,
+                * True for using the former correction
+        :param ndarray dark: dark noise image
+        :param lst_flat: numpy.Array or list of numpy.array which flat the lst_data.
+        :param IntegrationMethod method: IntegrationMethod instance or 3-tuple with (splitting, algorithm, implementation)
+        :param float normalization_factor: Value of a normalization monitor
+        :return: regrouped intensity and unit arrays
+        :rtype: Integrate2dResult
+        """
+        if "npt_horizontal" in kwargs:
```
kif, line 386: Since this block of code is duplicated twice, it would make sense to have it in a (private) method. One may need to update the locals dict ...

src/pyFAI/multi_geometry.py
```diff
@@ -342,3 +343,493 @@ def reset(self, collect_garbage=True):
...
+            oop_range = kwargs["vertical_unit_range"]
+
+        for fi in self.ais:
+            unit_ip, unit_oop = fi.parse_units(unit_ip=unit_ip, unit_oop=unit_oop,
+                                             sample_orientation=sample_orientation)
+
+            fi.reset_integrator(incident_angle=unit_ip.incident_angle,
+                                tilt_angle=unit_ip.tilt_angle,
+                                sample_orientation=unit_ip.sample_orientation)
+
```
kif, line 412: Always convert method to an instance of IntegrationMethod and then check the attributes.

EdgarGF93:
First, we merge #2324 and the code for this PR will be much lighter

EdgarGF93:
Needs also #2336

kif (approved):

PR #2341, opened 2024-11-19
[OCL] Explicitely disable fp64

### LINKED ISSUE #2339: GPU Acceleration on Newest Version Not Working
I have been trying to update a piece of code I wrote to run on an M1 Max MacBook Pro with 64 GB of RAM to make use of the gpu. In a previous version of pyFAI, this was accomplished easily using the method: ('full', 'csr', 'opencl', 'gpu'). Now, that method does not work. I tried going through the tutorial on the pyfai documentation (https://pyfai.readthedocs.io/en/stable/usage/tutorial/Parallelization/GPU-decompression.html) and the issue remains. 

```python
import sys, os, collections, struct, time
import numpy, pyFAI
import h5py, hdf5plugin
from matplotlib.pyplot import subplots
import bitshuffle
import pyopencl.array as cla
import silx
from silx.opencl import ocl
from silx.opencl.codec.bitshuffle_lz4 import BitshuffleLz4
start_time = time.time()
ocl
# Here we want to select the open CL device. 
target = (0,1)
det = pyFAI.detector_factory("eiger_4M")
shape = det.shape
dtype = numpy.dtype("uint32")
filename = "/tmp/big.h5"
nbins = 1000
cmp = hdf5plugin.Bitshuffle()
hdf5plugin.get_config().build_config
mem_bytes = os.sysconf('SC_PAGE_SIZE') * os.sysconf('SC_PHYS_PAGES')
print(f"Number of frames the computer can host in memory: {mem_bytes/(numpy.prod(shape)*dtype.itemsize):.3f}")
if os.environ.get('SLURM_MEM_PER_NODE'):
    print(f"Number of frames the computer can host in memory with SLURM restrictions: {int(os.environ['SLURM_MEM_PER_NODE'])*(1<<20)/(numpy.prod(shape)*dtype.itemsize):.3f}")
#The computer being limited to 64G of RAM, the number of frames actually possible is 3800.
nbframes = 4096 # slightly larger than the maximum achievable ! Such a dataset should not host in memory.

#Prepare a frame with little count so that it compresses well
geo = {"detector": det,
       "wavelength": 1e-10,
       "rot3":0} #work around a bug https://github.com/silx-kit/pyFAI/pull/1749
ai = pyFAI.load(geo)
omega = ai.solidAngleArray()
q = numpy.arange(15)
img = ai.calcfrom1d(q, 100/(1+q*q))
frame = numpy.random.poisson(img).astype(dtype)

# display the image
fig,ax = subplots()
ax.imshow(frame)
print("Performances of the different algorithms for azimuthal integration of Eiger 4M image on the CPU")
for algo in ("histogram", "csc", "csr"):
    print(f"Using algorithm {algo:10s}:", end=" ")
    %timeit ai.integrate1d(img, nbins, method=("full", algo, "cython"))
print("Performances of the different algorithms for azimuthal integration of Eiger 4M image on the GPU")
print(f"Using algorithm {algo:10s}:", end=" ")
%timeit ai.integrate1d(img, nbins, method=("full", algo, "opencl", target))
```

When I run this code, it fails when it uses the method: ('full', 'csr', 'opencl', target). 

The warning thrown is: WARNING:pyFAI.azimuthalIntegrator:MemoryError: falling back on default forward implementation

pyFAI can't recover from this and it seems to be stemming from not being able to reset the engine. It just gets stuck in an infinite wait loop.

kif:
Work aroung a bug in Apple Silicon GPU driver

Close #2339

### DIFF
--- a/doc/source/publications.rst
+++ b/doc/source/publications.rst
@@ -6,33 +6,39 @@
 Publications about pyFAI
 ========================
 
-* *PyFAI, a versatile library for azimuthal regrouping*,
-  J Kieffer & D Karkoulis;
+* *PyFAI, a versatile library for azimuthal regrouping*;
+  J Kieffer and D Karkoulis;
   **Journal of Physics: Conference Series** (2013) 425 (20), pp202012
   Initial publication where the usage of GPU is envisaged to overcome
   the speed limitation of azimuthal integration.
 
-* *PyFAI: a Python library for high performance azimuthal integration on GPU*
-  J Kieffer & J.P. Wright;
-  **Powder Diffraction** (2013) 28 (S2), pp339-350
-  Introduces the concept of look-up table to register the pixel distribution
+* *PyFAI: a Python library for high performance azimuthal integration on GPU*;
+  J Kieffer and J.P. Wright;
+  **Powder Diffraction** (2013) 28 (S2), pp339-350;
+  Introduces the concept of look-up table to register the pixel distribution.
 
 * *PyFAI: a Python library for high performance azimuthal integration on GPU*;
-   Jérôme Kieffer, Giannis Ashiotis
-   **PROC. OF THE 7th EUR. CONF. ON PYTHON IN SCIENCE** (EUROSCIPY 2014)
-   Presents the usage of sparse matrices to store the look-up table
+   Jérôme Kieffer and Giannis Ashiotis;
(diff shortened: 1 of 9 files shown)


DarioLewczyk (approved):
This is working on M1 Max Apple Silicon Device

Issue #2351, opened 2024-12-06
Worker.set_json_config failing when detector is a NexusDetector


t20100:
`Worker.set_json_config` fails with the following exception when config contains a `NexusDector`:

```
In [1]: from pyFAI.worker import Worker
In [2]: w = Worker()
In [3]: w.set_json_config("pyfai.json")
In [4]: w.get_json_config()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 w.get_json_config()

File lib/python3.11/site-packages/pyFAI/worker.py:628, in Worker.get_json_config(self)
    626 def get_json_config(self):
    627     """return configuration as a JSON string"""
--> 628     return json.dumps(self.get_config(), indent=2)

...

File lib/python3.11/json/encoder.py:180, in JSONEncoder.default(self, o)
    161 def default(self, o):
    162     """Implement this method in a subclass such that it returns
    163     a serializable object for ``o``, or calls the base implementation
    164     (to raise a ``TypeError``).
   (...)
    178 
    179     """
--> 180     raise TypeError(f'Object of type {o.__class__.__name__} '
    181                     f'is not JSON serializable')

TypeError: Object of type int32 is not JSON serializable
```

This looks to be due to the detector's `shape` which is a tuple of `numpy.int32` in this case.

BTW, the `shape` is saved by the worker but not by the GUI:
- https://github.com/silx-kit/pyFAI/blob/38250374b174f1caaeff9aca5c96462601651d61/src/pyFAI/worker.py#L590-L624
- https://github.com/silx-kit/pyFAI/blob/38250374b174f1caaeff9aca5c96462601651d61/src/pyFAI/gui/widgets/WorkerConfigurator.py#L239-L329

And it's not in the documented valid keys:
https://github.com/silx-kit/pyFAI/blob/38250374b174f1caaeff9aca5c96462601651d61/src/pyFAI/worker.py#L39-L78

So, should it be in the config?
Wouldn't that makes sense to have the definition of the config at a single place? e.g., a validator method and/or a TypedDict

kif:
close as fixed by #2352

PR #2352, opened 2024-12-06
NexusDetector: Fixed type on shape

t20100:
Not sure this is the right way to fix it, but this PR workaround #2351: by making sure the shape is a tuple of Python int, this PR fixes the export of the Worker config as a JSON file.

Related to #2351


kif (approved):
LGTM, thanks for spotting the bug & fixing it

PR #2374, opened 2025-01-07
rebin in 1d from 2d data 

kif:
close #2372


src/pyFAI/integrator/azimuthal.py
loichuder, line 1: Are the changes of this file related to the new function ? If so, in which way ?

kif: Not at all ... but the CI was blocked for an unknown reason.

PR #2380, opened 2025-01-17
Support for grazing-incidence exit angles and polar angle

EdgarGF93:
The units called previously exit angles should be "scattering_angles" in the classic way, because they don't depend on incident and tilt angles. Rotations are applied to q units.
Now, the unit called "exit angle" is used in the grazing-incidence community for the scattering angle with the origin in the horizon of the thin film, now dependant on the incident and tilt angles
The "exit_angles" support is especially useful for SAXS experiments


EdgarGF93:
image

EdgarGF93:
In a real example, we would integrate using the angle units:
```
%matplotlib widget
incident_angle = 0.0
tilt_angle = 0.0
res2d_0 = fi.integrate2d_grazing_incidence(data=arr, 
                                           unit_ip="scattering_angle_horz", unit_oop="exit_angle", 
                                           incident_angle=incident_angle, tilt_angle=tilt_angle,
                                          )

incident_angle = 0.2
tilt_angle = 0.0
res2d_1 = fi.integrate2d_grazing_incidence(data=arr, 
                                           unit_ip="scattering_angle_horz", unit_oop="exit_angle", 
                                           incident_angle=incident_angle, tilt_angle=tilt_angle,
                                          )

fig, axes = subplots(ncols=2)
plot2d(res2d_0, ax=axes[0])
plot2d(res2d_1, ax=axes[1])

```

image

EdgarGF93:
All these images will be generated in the tutorial notebook:

If we dont specify the units (so it will use qip_nm-1 and qoop_nm-1)
image
We see that our "correct" sample orientation is 6 (or 7, it only switches the x sign)

```
for so in range(1,9):
    ax_ = ax.ravel()[so-1]
    res = fi.integrate2d_grazing_incidence(data=data, sample_orientation=so,
                                       unit_ip="qtot_nm^-1",
                                       unit_oop="chigi_deg",
                                      )
```
image

So in this case, if we choose the correct sample orientation, the polar angle 0 will be associated always to the vertical axis of the sample (fiber axis or thin film normal vector)

EdgarGF93:
needs to fix https://github.com/silx-kit/pyFAI/issues/2389

EdgarGF93:
Needs to add tests for the different units with scales horizontal and vertical (non-regression)

kif (approved):

PR #2381, opened 2025-01-17
Use unit.name as key cache

EdgarGF93:
For two units called `scattering_angle_horz` and `scattering_angle_vert` ; the key to cache the arrays is the same for both units (`scattering`). 
Can we change it and use just the full unit.name?

### DIFF
--- a/src/pyFAI/geometry/core.py
+++ b/src/pyFAI/geometry/core.py
@@ -767,9 +767,6 @@ class Geometry(object):
             # unit = to_unit("r_m")
         key = space + "_corner"
 
-        if isinstance(unit, UnitFiber):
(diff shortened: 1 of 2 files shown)


kif:
If I understand correctly, the issue comes from the `_` in your unit name.
If we agree that the second half should never gave a `_` as separator, wouln't it be simpler to use:
`
space = "_".join(name.split("_")[:-1]))
`
instead of having a different behaviour for azimuthal and fiber units ?

EdgarGF93:
why we cannot use the whole unit.name as key?

EdgarGF93:
something like this?

kif (approved):
LGTM

PR #2392, opened 2025-01-23
[diffmap-view] fix shape key

EdgarGF93:


EdgarGF93:
Or we include the key "shape" in the Worker configuration, although it can be guess from poni

kif (approved):
LGTM 
I am considering many other changes in this part ...
I would rather keep the worker configuration in a centralized place.
So maybe it will change in the (near) future.

kif:
I foresee to keep track of the shape of the processed images in the worker. There is a PR on the topic to create this entry in the WorkerConfig dataclass but it is not yet used/populated.

Issue #2389, opened 2025-01-23
Wrong application of q_A-1 scale on the out-of-plane axis


EdgarGF93:
```
from pyFAI.gui.jupyter import plot2d, subplots
import matplotlib.pyplot as plt
from pyFAI.test.utilstest import UtilsTest
from pyFAI.units import to_unit
from pyFAI import load
import fabio

if __name__ == "__main__":
    data = fabio.open(UtilsTest.getimage("Y6.edf")).data
    ponifile = UtilsTest.getimage("LaB6_3.poni")
    ai = load(filename=ponifile)
    radial_range = [0,2]
    azimuth_range = [0,2]
    

    res_1 = ai.integrate2d(data=data, unit=(to_unit("chi_deg"), to_unit("q_A^-1")), npt_rad=1000, 
                         )
    res_2 = ai.integrate2d(data=data, unit=(to_unit("chi_deg"), to_unit("q_A^-1")), npt_rad=1000, 
                        azimuth_range=azimuth_range, radial_range=radial_range,
                         )
    
    res_3 = ai.integrate2d(data=data, unit=(to_unit("q_A^-1"), to_unit("chi_deg")), npt_rad=1000, 
                         )
    res_4 = ai.integrate2d(data=data, unit=(to_unit("q_A^-1"), to_unit("chi_deg")), npt_rad=1000, 
                        azimuth_range=azimuth_range, radial_range=radial_range,
                         )

    fig, ax = subplots(ncols=2, nrows=2)
    plot2d(res_1, ax=ax[0,0])
    plot2d(res_2, ax=ax[0,1])
    plot2d(res_3, ax=ax[1,0])
    plot2d(res_4, ax=ax[1,1])
    plt.show()
                              
```

![Image](https://github.com/user-attachments/assets/2a335921-2469-4e29-acd0-34a8c6f2a0c3)

EdgarGF93:
I saw it because in the grazing-incidence module, both axis are exchangeable.
If the "vertical" axis contains the unit q_A-1, the scale (*0.1) appears to be applied twice ?

EdgarGF93:
It hast to be unified here:
https://github.com/silx-kit/pyFAI/blob/ee734f16f5dcb17476654ff32b10262eaf21a84e/src/pyFAI/integrator/azimuthal.py#L721-L740

EdgarGF93:
Something like this (very quick) does work:

```
        if radial_range:
[...]

PR #2395, opened 2025-01-24
equivalence azimuth and radial units

EdgarGF93:
From #2389


EdgarGF93:
Now, the integrations with/without chi_discontinuity_at_zero, with degrees/radians:
image

kif (approved):

PR #2419, opened 2025-01-31
Diffmap

kif:
minor bug-fix

### DIFF
--- a/src/pyFAI/integrator/azimuthal.py
+++ b/src/pyFAI/integrator/azimuthal.py
@@ -30,7 +30,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "18/02/2025"
+__date__ = "19/02/2025"
 __status__ = "stable"
 __docformat__ = 'restructuredtext'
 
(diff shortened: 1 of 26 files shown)


kif:
This became a huge PR ... and not even sure it addresses everything.
Well all test passes, could you give me your point of view ?

loichuder:
> Well all test passes, could you give me your point of view ?

Sure. Could you give some context ? Like what bug are you trying to fix ? And how did you fix it ?

kif:
> Sure. Could you give some context ? Like what bug are you trying to fix ? And how did you fix it ?

The big issue is to be able to process data which were acquired binned (issue on ID27).
But this is strongly linked with the introduction of the dataclass
"WorkerConfig" for the configuration.

kif:
Here is the difference: images arrive with the shape 1500x1600 while the azimuthal integrator is setup for images of shape (3000x3200). Of course, this can be "guessed" but checking this every time an array could be of different size creates a lot of modifications a bit everywhere.

src/pyFAI/integrator/azimuthal.py
```diff
@@ -128,7 +128,7 @@ def integrate1d(self, data, npt, filename=None,
         method = self._normalize_method(method, dim=1, default=self.DEFAULT_METHOD_1D)
         assert method.dimension == 1
         unit = units.to_unit(unit)
-        empty = dummy if dummy is not None else self._empty
+        empty = numpy.float32(dummy) if dummy is not None else self._empty
```
loichuder, line 131: Why the cast?

loichuder (approved):

Issue #2430, opened 2025-02-10
[pyFAI-integrate] Crashes if we switch off integrate2d


EdgarGF93:

kif:
On the main branch or on 2025.01?

kif:
I believe this is fixed in the main branch.

PR #2438, opened 2025-02-14
Auto guess the binning from the reading of the first image

### LINKED ISSUE #2436: `diffmap` application does not support binned images
Need to deal with image size != detector max_shape. Several sub issues exist related to:
* Management of the size of the mask (static, dynamic)
* Size of the flat field image
* Size of the dark current image

kif:
close #2436 #2430


kif:
close #2420

PR #2441, opened 2025-02-17
fake data soft condition

EdgarGF93:


kif (changes requested):
2 lines before we calculate the FWHM from Cagliotti's law. 
Just keep all rings for `tth_min-3*sigma_min<t<tth_max+3*sigma_max`
The value 0.95 is completely arbitrary. 
```
sigma_min = (sigma2.min())**0.5
sigma_max = (sigma2.max())**0.5
```

kif (approved):
Looks good. Did you check it actually addresses the bug ?

EdgarGF93:
yes, it does correct the bug

Issue #2428, opened 2025-02-07
[FiberUnit] Units cannot be parsed ...


kif:
```
> pyFAI.units.to_unit(str(pyFAI.units.Q_OOP))

ERROR:pyFAI.units:Unable to recognize this type unit '
qoop_nm^-1
Incident_angle=0.0°
Tilt_angle=0.0°
Sample orientation=1
' of type <class 'str'>. Valid units are r_mm, r_m, 2th_deg, 2th_rad, q_nm^-1, q_A^-1, d_m, d_nm, d_A, d*2_A^-2, d*2_nm^-2, log10(q.m)_None, log(q.nm)_None, log(1+q.nm)_None, log(1+q.A)_None, arcsinh(q.nm)_None, arcsinh(q.A)_None, qx_nm^-1, qy_nm^-1, scattering_angle_vert_rad, scattering_angle_horz_rad, exit_angle_vert_rad, exit_angle_horz_rad, exit_angle_vert_deg, exit_angle_horz_deg, qxgi_nm^-1, qygi_nm^-1, qzgi_nm^-1, qip_nm^-1, qoop_nm^-1, qtot_nm^-1, qxgi_A^-1, qygi_A^-1, qzgi_A^-1, qip_A^-1, qoop_A^-1, qtot_A^-1, chi_rad, chi_deg, chigi_rad, chigi_deg

```

EdgarGF93:
#2442

PR #2442, opened 2025-02-18
fiberunit str method

EdgarGF93:


kif (approved):
Looks good. Does the arrays (qip&qoop) depend on the two parameter ? if so, one should consider cache invalidation ...

kif:
close #2428

PR #2450, opened 2025-02-20
[pilx] use dynamic masking for diffraction images

### LINKED ISSUE #2432: [pilx] Mask out the diffraction image
Eiger2 detector images feature very intense  masked pixels which makes the image less readable
From Thomas.
`silx -> image -> setMaskdata`

kif:
This allows to hide the intense gaps in Eiger images thus better auto- color scale.
close #2432

### DIFF
--- a/src/pyFAI/gui/pilx/MainWindow.py
+++ b/src/pyFAI/gui/pilx/MainWindow.py
@@ -33,7 +33,7 @@ __authors__ = ["Loïc Huder", "E. Gutierrez-Fernandez", "Jérôme Kieffer"]
 __contact__ = "loic.huder@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "27/01/2025"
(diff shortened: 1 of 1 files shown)


src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -279,10 +282,25 @@ def displayImageAtIndices(self, indices: ImageIndices):
...
+            detector_mask = detector.mask
+            if detector.shape != image.shape:
+                detector.guess_binning(image)
+                detector_mask = rebin_fct(detector_mask, detector.binning)
+            if mask_image is None:
+                mask_image = detector_mask
+            else:
+                numpy.logical_or(mask_image, detector_mask, out=mask_image)
+            detector.mask = mask_image
+            mask_image = detector.dynamic_mask(image)
```
loichuder, line 303: All these intertwined conditions make the code difficult to parse. I would suggest to move the highlighted part in a separate function (like `get_mask_file`) and use early returns. That will perhaps help a bit.

loichuder (approved):

PR #2455, opened 2025-02-24
Refactor getMask in pilx

loichuder:
Continuing work of https://github.com/silx-kit/pyFAI/pull/2450#discussion_r1963683041

@kif Do you have a file I could use to test ?


kif:
Look at those files:
http://www.silx.org/pub/pyFAI/pyFAI_UM_2025/

loichuder:
Are these raw data files ? There is no motor information inside.

How can I generate a diffmap file from those ?

PR #2468, opened 2025-03-12
Fakedata background

EdgarGF93:


kif:
Wouldn't it be simpler to have `Imin` and `Imax` as parameters, and modulate the intensities between the two ?
This would make:
```
signal = (Imax - Imin) * signal + Imin
```
Imin would be simpler to describe in the doc-string.

kif (approved):

PR #2472, opened 2025-03-13
numexpr on fiber

EdgarGF93:


kif (approved):
Normally, `numexpr` is a dependency of pyFAI ... you should not have to test for its presence. Beside this, LGTM

Issue #2474, opened 2025-03-14
Results of integration with `method = 'python'` are different between PyFAI 2025.03 and 2025.01


loichuder:
Take the following code (adapted from ewoksxrpd tests) that generates an image from a 1D pattern, and assert that the integrated pattern is the same:

```python
import numpy
import pyFAI.detectors
from pyFAI.integrator.azimuthal import AzimuthalIntegrator


def linspace(low_limit, high_limit, n):
    half_bin = (high_limit - low_limit) / (2 * n)
    return numpy.linspace(low_limit + half_bin, high_limit - half_bin, n)


x = linspace(0, 30, 1024)
y = numpy.zeros(x.size)
s = 0.5
for ufrac in [0.1, 0.4, 0.7]:
    A = 100
    u = x[0] + ufrac * (x[-1] - x[0])
    y += A * numpy.exp(-((x - u) ** 2) / (2 * s**2)) / (s * numpy.sqrt(2 * numpy.pi))


geometry = {
    "dist": 10e-2,  # 10 cm
    "poni1": 10e-2,  # 10 cm
    "poni2": 10e-2,  # 10 cm
    "rot1": numpy.radians(10),  # 10 deg
    "rot2": 0,  # 0 deg
    "rot3": 0,  # 0 deg
    "wavelength": 1e-10,
}


ai = AzimuthalIntegrator(
    detector=pyFAI.detectors.detector_factory("Pilatus1M"), **geometry
)
image = ai.calcfrom1d(
    x,
    y,
    dim1_unit="2th_deg",
)


integration_options = {
    "radial_range": (0, 30),
    "npt": 1024,
    "unit": "2th_deg",
    "method": "python",
}
result = ai.integrate1d(image, **integration_options)

atol = y.max() * 0.01
numpy.testing.assert_allclose(y, result.intensity, atol=atol)
```

**The assertion on the bottom breaks on PyFAI 2025.03 but goes fine on PyFAI 2025.01.**

Also, the assertion succeeded when using `method = "cython"`:

```python
integration_options = {
    "radial_range": (0, 30),
    "npt": 1024,
    "unit": "2th_deg",
    "method": "cython",
}
c_result = ai.integrate1d(image, **integration_options)
numpy.testing.assert_allclose(y, c_result.intensity, atol=atol)
```

kif:
[...]

PR #2475, opened 2025-03-15
Update python histogram

kif:
So that the empty bins gets filled with the empty value instead of the dummy value.


kif:
close #2474

loichuder:
:+1:

PR #2497, opened 2025-04-03
fix some fiber units

EdgarGF93:


EdgarGF93:
I also realized that there are some Angstrom fiber units without numexpr

src/pyFAI/units.py
```diff
@@ -1189,13 +1193,15 @@ def eq_chi_gi(x, y, z, wavelength, incident_angle=0.0, tilt_angle=0.0, sample_or
...
 register_azimuthal_fiber_unit(name="chigi_rad",
+                              short_name="chigirad",
                               scale=1.0,
                               label=r"Polar angle $\chi$ ($rad$)",
                               formula=formula_chi_gi,
                               equation=eq_chi_gi,
                               positive=False,
                               period=2.*pi)
 register_azimuthal_fiber_unit(name="chigi_deg",
+                              short_name="chigideg",
```
kif, line 1204: If 2 units differ only by the scale factor, they should have the same short_name ...

kif (approved):

PR #2501, opened 2025-04-04
diffmap dataclass

### LINKED ISSUE #2498: [DiffMap] Implement configuration using dataclass
Similar to `integration_config.py`

kif:
close #2498


kif:
At this point, the dataclass is ready, but not used at all. 
The second stage of the work consists in wiring it in diffmap and diffmap_widget source files. which should get simpler (and more reliable).

kif:
Maybe it will be easier to review the creation of the dataclass (at this stage) and put into another PR the usage of the dataclass. 
@loichuder would you have time to review this PR or shall I ask someone else ?

I just ran a code coverage test and `pyFAI/io/diffmap_config.py` is at 70%, better than the average pyFAI code.

src/pyFAI/io/diffmap_config.py
```diff
@@ -0,0 +1,347 @@
...
+
+    @classmethod
+    def from_tuple(cls, tpl):
+        return cls(*tpl)
+
+    def is_hdf5(self):
+        """Return True if the object is hdf5"""
+        if self.h5 is None:
+            self.h5 = is_hdf5(self.path)
+        return bool(self.h5)
```
loichuder, line 114: Is there really a reason to cache `self.h5` ? Why not making it a computed `property` ?

src/pyFAI/io/diffmap_config.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+class ListDataSet(list):
+
+    def commonroot(self):
+        """
+        :return: common directory
+        """
+        ll = [j.path.split(os.sep) for j in self]
+        common = os.path.commonprefix(ll)
```
loichuder, line 127: If you wish to return the common directory, you might want to use [commonpath](https://docs.python.org/3/library/os.path.html#os.path.commonpath) instead.

src/pyFAI/io/diffmap_config.py
```diff
@@ -0,0 +1,347 @@
...
+        "nbpt_fast": "fast_motor.points",
+        "nbpt_slow": "slow_motor.points",
+        "fast_motor_points": "fast_motor.points",
+        "slow_motor_points": "slow_motor.points",
+        "fast_motor_name": "fast_motor.name",
+        "slow_motor_name": "slow_motor.name",
+        "fast_motor_range": "fast_motor.start|stop",
+        "slow_motor_range": "slow_motor.start|stop",
+        "npt_rad": "ai.nbpt_rad",
+        "npt_azim": "ai.nbpt_rad",
```
loichuder, line 203:
```suggestion
        "npt_azim": "ai.nbpt_azim",
```

src/pyFAI/io/diffmap_config.py
```diff
@@ -0,0 +1,347 @@
...
+    @classmethod
+    def from_file(cls, filename: str):
+        """load the content of a JSON file and provide a dataclass instance"""
+        with open(filename, "r") as f:
+            dico = json.loads(f.read())
+        return cls.from_dict(dico, inplace=True)
+
+    # Compatibility layer !
+    @property
+    def npt_fast(self):
```
loichuder, line 294: Are there deprecation warnings raised by these ?

src/pyFAI/test/test_io_diffmap_config.py
```diff
@@ -0,0 +1,198 @@
...
+""" # data obtained with pyFAI 2025.
+
+class TestDiffmapConfig(unittest.TestCase):
+    """Test diffmap config
+    """
+    @classmethod
+    def setUpClass(cls):
+        cls.inp = json.loads(test_data)
+    @classmethod
+
```
loichuder, line 160:
```suggestion
```

src/pyFAI/test/test_io_diffmap_config.py
```diff
@@ -0,0 +1,198 @@
...
+            else:
+                self.assertTrue(isinstance(value, field.type), msg)
+        #just a few test to validate the parsing...
+        self.assertEqual(parsed.output_file, self.inp["output_file"])
+        self.assertEqual(parsed.diffmap_config_version, CURRENT_VERSION)
+        self.assertEqual(parsed.experiment_title, self.inp["experiment_title"])
+        self.assertEqual(parsed.offset, self.inp["offset"])
+        self.assertEqual(parsed.zigzag_scan, self.inp["zigzag_scan"])
+
+    def test_consistant(self):
```
loichuder, line 180:
```suggestion
    def test_consistency(self):
```

src/pyFAI/diffmap.py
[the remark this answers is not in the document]
kif: Unfortunately not all ...GH shows modifications from #2499 which have already been merged into the main branch. but it is a lot less.

src/pyFAI/io/diffmap_config.py
kif: Good remark ... since I just "dataclassed" an existing class
https://github.com/silx-kit/pyFAI/blob/55d53cf195d9600689e3a29f0626b3d1b6bb917b/src/pyFAI/gui/utils/tree.py#L147
I did not re-think the way it was working.

Actually this object is strange since `self.h5` is now the path of the hdf5 file and contains no reference to the internal path. I opened an issue to think specifically about this issue: #2503

src/pyFAI/io/diffmap_config.py
kif: Not yet ... but it is the place where to put them.

Issue #2512, opened 2025-04-30
[pilx] compatibility with zigzag scans


kif:
The best would be to store an array called `map_ptr` of the shape of the map with the frame index in them.

* [x] diffmap should create a `map_ptr` array of integers which are pointers to the frame index.
* [x] `map_ptr` can be created from the motor position from the bliss master file
* [x] pilx shoud use the `map_ptr` to determine the index of the frame

PR #2524, opened 2025-05-14
Implement the storage of the `map_ptr` in 

kif:
related to #2512

### DIFF
--- a/src/pyFAI/gui/pilx/MainWindow.py
+++ b/src/pyFAI/gui/pilx/MainWindow.py
@@ -33,10 +33,11 @@ __authors__ = ["Loïc Huder", "E. Gutierrez-Fernandez", "Jérôme Kieffer"]
 __contact__ = "loic.huder@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "20/02/2025"
+__date__ = "14/05/2025"
 __status__ = "development"
 
 from typing import Tuple
+from string import digits
 import json
 import h5py
 import logging
@@ -117,6 +118,9 @@ class MainWindow(qt.QMainWindow):
         self._background_point = None
         self.worker_config = None
 
+        # declaration of instance variables
+        self._map_ptr = None # This is the map of the indices of input frame
+
     def initData(self,
                  file_name: str,
                  dataset_path: str="/entry_0000/measurement/images_0001",
@@ -132,6 +136,10 @@ class MainWindow(qt.QMainWindow):
         with h5py.File(self._file_name, "r") as h5file:
             nxprocess = h5file[self._nxprocess_path]
             map_data = get_dataset(nxprocess, "result/intensity")[()].sum(axis=-1)
+            try:
+                self._map_ptr = get_dataset(nxprocess, "result/map_ptr")[()]
+            except Exception:
+                self._map_ptr = numpy.arange()
             try:
                 slow = get_dataset(nxprocess, "result/slow")
             except Exception:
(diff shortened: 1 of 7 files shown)


kif:
Image indexes are now retrieved from a `map_ptr` array, 
* `diffmap` has been modified to build such an array
* `pilx` has been modified to benefit from such an array  (with a fallback to the previous formula)
* pyFAI.io.difffmap_config has now a function that guesses the parameter range.

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -117,6 +118,9 @@ def __init__(self) -> None:
         self._background_point = None
         self.worker_config = None
 
+        # declaration of instance variables
+        self._map_ptr = None # This is the map of the indices of input frame
```
loichuder, line 122: The fact that a comment is needed makes me think that `map_ptr` is perhaps not explicit enough as a name.

`frame_indices` ? `position_to_frame`?

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -132,6 +136,10 @@ def initData(self,
         with h5py.File(self._file_name, "r") as h5file:
             nxprocess = h5file[self._nxprocess_path]
             map_data = get_dataset(nxprocess, "result/intensity")[()].sum(axis=-1)
+            try:
+                self._map_ptr = get_dataset(nxprocess, "result/map_ptr")[()]
+            except Exception:
```
loichuder, line 141: `Exception` can probably be more specific ? `KeyError` ?

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -132,6 +136,10 @@ def initData(self,
...
+                self._map_ptr = numpy.arange()
```
loichuder, line 142: I would suggest to log a warning here, saying that the relationship between map positions and frame index could be wrong.

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -164,15 +172,15 @@ def initData(self,
...
-                if char.isdigit():
-                    cnt += 1
-                else:
-                    break
+            try:
+                self._map_ptr = get_dataset(nxprocess, "result/map_ptr")[()]
+            except Exception:
+                self._map_ptr = numpy.arange(fast.size*slow.size)
+                self._map_ptr += self._offset
+                self._map_ptr.shape = (slow.size, fast.size)
```
loichuder, line 180: Same comments as above

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -282,7 +290,10 @@ def displayImageAtIndices(self, indices: ImageIndices):
         with h5py.File(self._file_name, "r") as h5file:
             nxprocess = h5file[self._nxprocess_path]
             map_shape = get_dataset(nxprocess, "result/intensity").shape
-            image_index = row * map_shape[1] + col + self._offset
+            if self._map_ptr is not None:
```
loichuder, line 293: According to above, `self._map_ptr` should never be `None` here, right ?

src/pyFAI/gui/pilx/MainWindow.py
kif: To me it is a `map of pointers` with minimal documentation about what it is.
Since I dislike formatter which are wrapping lines (making code unreadable IMHO), keeping variable names short limits such wrapping, thus the short variable names.

src/pyFAI/gui/pilx/MainWindow.py
kif: Probably, the original code is sometimes checking against both of them. I will migrate them all to KeyError if possible

src/pyFAI/gui/pilx/MainWindow.py
kif: Can you guarantee `initData` is always run before-hand ?

src/pyFAI/gui/pilx/MainWindow.py
loichuder: > keeping variable names short limits such wrapping

There is no formatter run here so I don't understand your justification.

Anyway, if the variable name is explicit enough for you, good. My comments are and stay suggestions.

src/pyFAI/gui/pilx/MainWindow.py
loichuder: Yes: if `initData`, was not run, `self._file_name` will be `None` and the function will return at line 286 just above.

PR #2532, opened 2025-05-28
use relative path for default NeXus attribute

kif:
The rational is to avoid breaking display when using external links


axelboc (approved):
Thanks!

PR #2533, opened 2025-06-04
offer several type of Qt binding

kif:


kif:
@picca, just to let you know we are gradually migrating to Qt6, mostly PySide6, but PyQt6 is also an option. The bugs are not the same with those bindings

PR #2543, opened 2025-06-15
Grazing incidence improvements

EdgarGF93:
- [x] Integrate2dFiber container
- [x] Integrate1dFiber container
- [x] Missing wedge method (histogram2d + smooth algorithm)
- [x] Warnings to use the pixel splitting + missing wedge mask
- [x] Fix tutorial

### DIFF
--- a/src/pyFAI/test/test_fiber_integrator.py
+++ b/src/pyFAI/test/test_fiber_integrator.py
@@ -311,17 +311,17 @@ class TestFiberIntegrator(unittest.TestCase):
         res1d_vertical = self.fi.integrate1d_grazing_incidence(data=self.data,
                                                                npt_oop=100, npt_ip=500,
                                                                vertical_integration=True)
-
-        self.assertEqual(abs(res1d_vertical_ref.radial - res1d_vertical.radial).max(), 0)
-        self.assertEqual(abs(res1d_vertical_ref.intensity - res1d_vertical.intensity).max(), 0)
+        
+        numpy.allclose(res1d_vertical_ref.radial, res1d_vertical.radial)
+        numpy.allclose(res1d_vertical_ref.intensity, res1d_vertical.intensity)
 
         res1d_horizontal_ref = self.fi.integrate1d_grazing_incidence(data=self.data, npt_ip=100, vertical_integration=False)
         res1d_horizontal = self.fi.integrate1d_grazing_incidence(data=self.data,
                                                                npt_oop=500, npt_ip=100,
                                                                vertical_integration=False)
-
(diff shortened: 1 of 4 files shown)


EdgarGF93:
Now, there is the possibility of masking the missing wedge on a pixel-splited result:
Example, oversampling on a 2d integration:
image

Extreme oversampling on a little patch of the image:
image

So, while the pixel splited integration is wrong (there is no missing wedge), and the non-splited falls into oversampling, we can patch the first one:
image

EdgarGF93:
For the moment, the function to find a threshold to mask the missing wedge is quite naive, but works pretty nice with these couple of examples

src/pyFAI/test/test_fiber_integrator.py
```diff
@@ -312,16 +312,16 @@ def test_integrate1d_defaults(self):
                                                                npt_oop=100, npt_ip=500,
                                                                vertical_integration=True)
 
-        self.assertEqual(abs(res1d_vertical_ref.radial - res1d_vertical.radial).max(), 0)
-        self.assertEqual(abs(res1d_vertical_ref.intensity - res1d_vertical.intensity).max(), 0)
+        numpy.allclose(res1d_vertical_ref.radial, res1d_vertical.radial)
```
kif: You are note making a test (assert) to validate the result is correct ... Here the test always passes, the result of `numpy.allclose` can just be False.

src/pyFAI/test/test_fiber_integrator.py
```diff
@@ -312,16 +312,16 @@ def test_integrate1d_defaults(self):
...
 
         res1d_horizontal_ref = self.fi.integrate1d_grazing_incidence(data=self.data, npt_ip=100, vertical_integration=False)
         res1d_horizontal = self.fi.integrate1d_grazing_incidence(data=self.data,
                                                                npt_oop=500, npt_ip=100,
                                                                vertical_integration=False)
 
-        self.assertEqual(abs(res1d_horizontal_ref.radial - res1d_horizontal.radial).max(), 0)
-        self.assertEqual(abs(res1d_horizontal_ref.intensity - res1d_horizontal.intensity).max(), 0)
+        numpy.allclose(res1d_horizontal_ref.radial, res1d_horizontal.radial)
+        numpy.allclose(res1d_horizontal_ref.intensity, res1d_horizontal.intensity)
```
kif, line 325: Same here.
Use `self.assertTrue(numpy.allclose(...))`

EdgarGF93: right, sorry for that

Issue #2527, opened 2025-05-16
[pilx] lots of warnings with PySide6 backend


kif:
```
(py313) lintaillefer:BM29/Mesh-20250403/chi_chan_seu_channel_chip_mesh_scan_channel_chip_0002 % pyFAI-diffmap-view scan17_zigzag.h5 
/users/kieffer/.venv/py313/lib/python3.13/site-packages/pyFAI/gui/pilx/HorizontalRangeROI.py:50: RuntimeWarning: 

*** Sort Warning ***
Signals and slots in QMetaObject 'HorizontalRangeROI' are not ordered correctly, this may lead to issues.
1  Slot   _editingStarted()
2  Slot   _editingFinished()
3! Signal sigRangeChanged()

  self.sigEditingFinished.connect(self.sigRangeChanged)
qt.core.qobject.connect: QObject::connect: No such signal HorizontalRangeROI::sigRangeChanged()
/users/kieffer/.venv/py313/lib/python3.13/site-packages/silx/gui/plot/items/roi.py:1415: RuntimeWarning: 

*** Sort Warning ***
Signals and slots in QMetaObject 'HorizontalRangeROI' are not ordered correctly, this may lead to issues.
1  Slot   _editingStarted()
2  Slot   _editingFinished()
3! Signal sigRangeChanged()
4! Slot   _minPositionChanged(PyObject)

  self._markerMax.sigItemChanged.connect(self._maxPositionChanged)
/users/kieffer/.venv/py313/lib/python3.13/site-packages/silx/gui/plot/items/roi.py:1416: RuntimeWarning: 

*** Sort Warning ***
Signals and slots in QMetaObject 'HorizontalRangeROI' are not ordered correctly, this may lead to issues.
1  Slot   _editingStarted()
2  Slot   _editingFinished()
3! Signal sigRangeChanged()
4! Slot   _minPositionChanged(PyObject)
5! Slot   _maxPositionChanged(PyObject)

  self._markerCen.sigItemChanged.connect(self._cenPositionChanged)
/users/kieffer/.venv/py313/lib/python3.13/site-packages/silx/gui/plot/items/core.py:432: RuntimeWarning: 

*** Sort Warning ***
Signals and slots in QMetaObject 'HorizontalRangeROI' are not ordered correctly, this may lead to issues.
1  Slot   _editingStarted()
2  Slot   _editingFinished()
3! Signal sigRangeChanged()
4! Slot   _minPositionChanged(PyObject)
[...]

PR #2545, opened 2025-06-18
Make `pilx` compatible with mesh-scan files produced by dahu on bm29

### LINKED ISSUE #2513: [Pile] compatibility with BR29.Mesh files

kif:
Related to 
https://github.com/kif/dahu/issues/91
close #2513

### DIFF
--- a/src/pyFAI/app/pilx.py
+++ b/src/pyFAI/app/pilx.py
@@ -40,7 +40,7 @@ import sys
 import argparse
 import logging
 from ..gui.pilx.MainWindow import MainWindow
-from .. import version as pyFAI_version, __date__ as pyFAI_date
+from .. import version as pyFAI_version, date as pyFAI_date
 from ..io.nexus import Nexus
 
 logger = logging.getLogger(__file__)
--- a/src/pyFAI/app/pilx.py
+++ b/src/pyFAI/app/pilx.py
@@ -32,15 +32,47 @@ __author__ = "Loïc Huder"
 __contact__ = "loic.huder@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "17/06/2025"
+__date__ = "18/06/2025"
 __status__ = "development"
 
 from silx.gui import qt
 import sys
 import argparse
+import logging
 from ..gui.pilx.MainWindow import MainWindow
 from .. import version as pyFAI_version, __date__ as pyFAI_date
+from ..io.nexus import Nexus
 
+logger = logging.getLogger(__file__)
+
+def guess_file_type(filename, default="diffmap"):
+    """return the type of file to set the proper reader"""
+
+    def read_str(entry, key):
+        try:
+            raw = entry[key][()]
+        except:
+            res = ""
+        else:
+            if isinstance(raw, bytes):
+                raw = raw.decode()
+            res = raw.lower()
+        return res
+
+    with Nexus(filename, "r") as nxs:
+        entry = nxs.get_entries()[0]
+        program_name = read_str(entry, "program_name")
+        title = read_str(entry,"title")
+
+    if program_name == "bm29.mesh" or title == "biosaxs mesh experiment":
+        filetype = "bm29"
+    elif program_name=="pyfai" or title == "diff_map":
+        filetype = "diffmap"
+    else:
+        logger.warning(f"Unable to identify file: '{filename}' with program_name: '{program_name}' and title: '{title}'")
+        filetype = ""
+    logger.info(f"detected file type: {filetype}")
+    return filetype or default
 
 def main(args=None):
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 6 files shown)


src/pyFAI/app/pilx.py
```diff
@@ -51,16 +51,25 @@ def main(args=None):
...
     version = f"pyFAI-diffmap-view version {pyFAI_version}: {pyFAI_date}"
     parser.add_argument("-V", "--version", action='version', version=version)
 
     options = parser.parse_args(args)
+    if options.bm29:
+        data_path = "/entry_0000/1_mesh/sources/images_0000"
+        nxprocess_path = "/entry_0000/1_mesh"
+    else:
+        nxprocess_path = options.nxprocess_path
+        data_path = options.data_path
```
loichuder: Are we sure we want to put ESRF specific code in pyFAI ?

Could we not providing a util at the beamline to call `pyFAI-diffmap` that sets the appropriate `-p`  and `-d` instead ?

src/pyFAI/gui/pilx/utils.py
```diff
@@ -77,6 +77,54 @@ def get_dataset(parent: h5py.Group | h5py.File, path: str) -> h5py.Dataset:
     return dset
 
 
+def get_NXdata(parent: h5py.Group | h5py.File,
+               path: str = None):
```
loichuder:
```suggestion
               path: str | None = None):
```

For the other functions below as well

src/pyFAI/gui/pilx/utils.py
```diff
@@ -77,6 +77,54 @@ def get_dataset(parent: h5py.Group | h5py.File, path: str) -> h5py.Dataset:
...
+
+def get_signal_dataset(parent: h5py.Group | h5py.File,
+                       path: str=None,
+                       default: str="intensity") -> h5py.Dataset:
+    "Read the `signal` dataset associated to a NXdata, if no signal is provided, use the default one"
+    nxdata = get_NXdata(parent, path)
+    attrs = nxdata.attrs
+    dset = nxdata[attrs.get("signal",default)]
+    if not isinstance(dset, h5py.Dataset):
+        raise RuntimeError("dataset is not a `h5py.Dataset` instance")
```
loichuder: I'd say that this is a `TypeError` rather than a `RuntimeError` but ok.

src/pyFAI/gui/pilx/utils.py
```diff
@@ -77,6 +77,54 @@ def get_dataset(parent: h5py.Group | h5py.File, path: str) -> h5py.Dataset:
...
+    return dset
+
+def get_axes_dataset(parent: h5py.Group | h5py.File,
+                     path: str=None,
+                     dim: int=0,
+                     default: str="intensity") -> h5py.Dataset:
```
loichuder: Are you sure you want the default axes to be `"intensity"` ?

src/pyFAI/gui/pilx/utils.py
```diff
@@ -77,6 +77,54 @@ def get_dataset(parent: h5py.Group | h5py.File, path: str) -> h5py.Dataset:
...
+        dset_name = default
+    else:
+        if dim<0:
+            dim += len(axes)
+        if dim>=len(axes):
+            dset_name = default
+        else:
+            dset_name = axes[dim]
+    if dset_name not in nxdata:
+        raise RuntimeError("dataset `{dset_name}` does not exist in NXdata `{nxdata.name}`.")
```
loichuder:
```suggestion
        raise RuntimeError(f"dataset `{dset_name}` does not exist in NXdata `{nxdata.name}`.")
```

Missing `f` for the formatting string

src/pyFAI/app/pilx.py
kif: I re-worked a bit this part ... if the file comes from `bm29.mesh` the nxprocess_path is now set accordingly but does not affect the default behaviour.

loichuder: Still, I think it can be confusing to overwrite user inputs like `nxprocess_path`  and `data_path`. 

If I give a ` data_path` but `reader` detects `bm29` or `diffmap`, my `data_path` will be discarded.

But perhaps it is ok since it is not used ? :shrug:

loichuder (approved):

kif:
Hi Loic,
You reviewed faster than I polished the code ...

The main modifications are:
* Fix the  issue #2527 related to pyside6. Your code duplicated a signal from one of the parent class in silx. I removed the signal you introduced and switched over to the original one. The warnings have disappeared after the modifications in `src/pyFAI/gui/pilx/HorizontalRangeROI.py`. Since this solution was found with @t20100 maybe Thomas has some opinion on it. This is the modification I am the least confident with.
* Those 3D signal dataset can be seen as 2d map of spectra or 1D stack of images. Both are valid and depending on the stage of the processing. `pilx` has always been brittle regarding this aspect. This PR uses the `interpretation` attribute to guess if we have a map of spectra or  a stack of images and adapts the reading of the data (and all minor processing accordingly).
* The rest is minor modification, I am still not a fan of typing, so thanks for fixing them when you notice, but to me docstrings are more important than typing. I know we disagree.

src/pyFAI/app/pilx.py
kif: Probably if the option is provided, it should overwrite the default guessing ... you are right.

Issue #2555, opened 2025-07-02
[Crystallography]: Validate all 230 space groups for extinction rules


kif:
Despite the code from `pyFAI.crystallography.space_groups` has been generated by AI, it is full of errors in its implementation (sometimes the space-group number does not match with the name !). 

There is a non regression test in `pyFAI.test.test_calibrant` which checks the output of **validated** methods is correct. 

The work to do is the following:
Pick any of the method named `group*` and append `validated` to its docstring.
Run the tests:
```
python3 run_tests.py pyFAI.test.test_calibrant.TestReflection
```
It should fail with a list of hkl which are not correct. 
The [international tables of crystallography](https://it.iucr.org/Ac/ch2o3v0001/)  are available online, each space-group has some *Reflection conditions* which needs to match what is implemented in Python.

kif:
After the merge of #2554, space-group n° 1 to 19 have been validated but most of the other still contain errors.

For example, flag the method `group20_c2221` as *validated* in the docstring, when running the test, which fails as expected:
```
======================================================================
FAIL: test_code (pyFAI.test.test_calibrant.TestReflection.test_code)
Checks the class has no issue and that validated methods are actually correct !
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/users/kieffer/workspace-400/pyFAI/build/lib/python3.13/site-packages/pyFAI/test/test_calibrant.py", line 329, in test_code
    raise AssertionError(f"Space group {name} did not validate against xrayutilities")
AssertionError: Space group group20_c2221 did not validate against xrayutilities

Stdout:
[...]

PR #2565, opened 2025-07-07
[Crystallography] Extinction rules for differents space groups

gudlot:
This PR is a work in progress for adding reflection support as discussed in [#2555](https://github.com/silx-kit/pyFAI/issues/2555).

Feedback is welcome during development. I’ll mark as “Ready for review” when it's finished.

### DIFF
--- a/src/pyFAI/crystallography/space_groups.py
+++ b/src/pyFAI/crystallography/space_groups.py
@@ -152,11 +152,50 @@ class ReflectionCondition:
         return True
 
     @staticmethod
-    def group15_c2c_b(h, k, l):
-        """Space group 15: C 1 2/c 1(unique axis b). C-centering: (h + k) even. (0 k 0): k even only. validated"""
+    def group15_C2c(h: int, k: int, l: int) -> bool:
+        """
+        Space group 15: C 2/c. Monoclinic, unique axis b.
+
+        Valid reflections must satisfy:
+        - General hkl:           h + k even
+        - h0l (k = 0):           h, l even
+        - 0kl (h = 0):           k even
+        - hk0 (l = 0):           h + k even
+        - 0k0 (h = 0, l = 0):    k even
+        - h00 (k = 0, l = 0):    h even
+        - 00l (h = 0, k = 0):    l even
+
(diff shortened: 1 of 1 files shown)


kif:
Well the file is supposed to be cached in the `/tmp` directory:
`os.listdir(pyFAI.test.utilstest.UtilsTest.resources.data_home)`

Those files are all available on:
http://www.silx.org/pub/pyFAI/testimages/

I have very often tests failing in the CI because of files that could not be downloaded, there is a mechanism that ensures the checksum of the file is correct (which allows to upgrade the file when needed, enforcing the testing machine to re-download the file) ...

Moreover, there is nothing related to the skipping of tests in  https://github.com/silx-kit/silx/blob/main/src/silx/utils/ExternalResources.py

I don't know if this is a feature introduced by pytest ... but it could be an extra reason not using it :trollface: 

I just checked, with the file missing, the test takes 300ms (including the download) while the second run takes only 80ms.

kif:
Hi Gudrun,
Your code looks correct to me, I just checked in the international tables. BTW, I realized those tables are not freely available on the internet, as they are from inside ESRF.
This SG deserves specific investigation, could be there is an issue in `xrayutilities` or elsewhere.
Thanks for the huge job you already achieved.
Jerome

kif (approved):
Very good documentation.

gudlot:
### Next Space Groups Implemented

Hi @kif 

Here are the next space groups. I mostly followed the ITC definitions.  
In cases where the `xrayutilities` tests failed, I consulted the rules listed at [http://img.chem.ucl.ac.uk/sgp/large/sgp.htm](http://img.chem.ucl.ac.uk/sgp/large/sgp.htm) and adapted the implementations accordingly.  
With those adjustments, the tests against `xrayutilities` passed.

SG 110 was particularly tricky. If you have time, maybe you could take a look at it later.  
Currently, all tests pass.

Best,  
Gudrun

gudlot:
I found the solutions for SG167, SG161, on offer two solutions for SG167. All tests green AFAIK.

kif (approved):
Impressive work. Looks great !

PR #2570, opened 2025-07-09
Refactor the way xrpd 1d/2d are generated

### LINKED ISSUE #2563: [Crystallography] Correct the way the powder diffraction pattern gets generated
Several issues: 

* [ ] Intensity calculation to be corrected and based on the intensity of the 
* [ ] The width of the peak should depend on the peak position rather than on the bin position (unexpected asymmetry)
* [ ] Replace Caglioti's lay with the one from https://journals.iucr.org/paper?S2053273321007506

Partially implemented in https://github.com/silx-kit/pyFAI/pull/2553
Sub task of #2558

kif:
* Provide several resolution functions
* Break some tests which are relying on fake images
close #2563

### DIFF
--- a/src/pyFAI/crystallography/calibrant.py
+++ b/src/pyFAI/crystallography/calibrant.py
@@ -509,7 +509,7 @@ class Calibrant:
 
         # background can be an array
         if isinstance(background, Iterable):
-            background = numpy.interp(tth_deg,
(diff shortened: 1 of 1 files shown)


kif:
@EdgarGF93, this PR modifies the way powder diffraction patterns are calculated thus one of your test failing ... could you investigate ?

PR #2580, opened 2025-07-16
parallax with full pixel splitting issue

### LINKED ISSUE #2579: [parallax] The Cython fast path does not take into account parallax corrections.
Indeed, the position matrix comes directly out of detector and is directly used in some cython code `_geometry.calc_rad_azim` without going via the `_correct_parallax`  function. (`pyFAI.gerometry.core` L804)

If p1 and p2 could be directly corrected for parallax calculation ...
Not that simple !

kif:
This is mostly related to the cython fast-path.

close #2579


src/pyFAI/geometry/core.py
```diff
@@ -308,7 +315,52 @@ def _correct_parallax(self, d1, d2, p1, p2):
...
+            p1 += delta1
+            p2 += delta2
+        return delta1, delta2
+
+    def _calc_cartesian_positions(self,
+                                  d1:numpy.ndarray,
+                                  d2:numpy.ndarray,
+                                  poni1:float|None=None,
+                                  poni2:float|None=None,
+                                  do_parallax:bool=True):
```
loichuder:
```suggestion
                                  do_parallax:bool=True) -> tuple[numpy.ndarray, numpy.ndarray, numpy.ndarray]:
```

src/pyFAI/geometry/core.py
```diff
@@ -803,14 +864,20 @@ def corner_array(self, shape=None, unit=None, use_cython=True, scale=True):
...
                                                           self.rot1, self.rot2, self.rot3,
                                                           p1, p2, p3,
                                                           space, self._wavelength,
                                                           orientation=self.detector.orientation,
                                                           chi_discontinuity_at_pi=self.chiDiscAtPi)
-                        except KeyError:
-                            logger.warning("No fast path for space: %s", space)
+                        except KeyError as err:
+                            if "Parallax" in str(err):
+                                logger.warning(err)
```
loichuder: Might be cleaner to raise another error type rather than `KeyError` instead of testing of the error message.

Could be a custom error type (e.g. `ParallaxNotImplemented`) if you fear to catch an error from the same type that comes from another source

src/pyFAI/geometry/core.py
```diff
@@ -1104,7 +1171,7 @@ def array_from_unit(self, shape=None, typ="center", unit=units.TTH, scale=True):
             unit2 = None
             unit = to_unit(unit)
         meth_name = unit.get(typ)
-        if meth_name and meth_name in dir(Geometry):
+        if False: #meth_name and meth_name in dir(Geometry):
```
loichuder: Is this voluntary?

src/pyFAI/geometry/core.py
```diff
@@ -318,22 +370,22 @@ def _calc_cartesian_positions(self, d1, d2, poni1=None, poni2=None, do_parallax=
         :param d2: ndarray of dimention 1/2 containing the X pixel positions
         :param poni1: value in the Y direction of the poni coordinate (meter)
         :param poni2: value in the X direction of the poni coordinate (meter)
-        :param do_parallax: position should be corrected for parallax effect
+        :param do_parallax: set to True to correct for the parallax effect
         :return: 2-arrays of same shape as d1 & d2 with the position in meters
```
loichuder: This actually returns 3 arrays?

loichuder (approved):

src/pyFAI/geometry/core.py
kif: yes it is ... it a obfuscated way to remind me I should clean up.
Removing this bit of code is straight forwards ... but most of the called function in the `True` section have to be cleaned up as well.

Issue #2590, opened 2025-08-20
`chi` array is always calculated through `numexpr`, never via `numpy`


kif:
There is no equation implemented in numpy, no reference implemention to validate against.

PR #2591, opened 2025-08-20
Provide numpy implementation of azimuthal units

kif:
plus:
- ruff format the 2 files
- optimize slightly the execution speed (`numpy` is much slower than `python` when working on scalars)


kif:
close #2590, non regression test already in place.

EdgarGF93 (approved):

PR #2596, opened 2025-09-02
handle semi-defined ranges

### LINKED ISSUE #2586: radial_range: either min or max could be infinite
When I use the worker with integration option `radial_range_min=1` without specifying `radial_range_max`, the q values are all `NaN`.

I did not see this issue before and here is why:

## `pyFAI>=2025.1.0`

The worker gets `radial_range` from the `pyFAI.io.integration_config.WorkerConfig` object

https://github.com/silx-kit/pyFAI/blob/55d53cf195d9600689e3a29f0626b3d1b6bb917b/src/pyFAI/worker.py#L495

When only `radial_range_min=n` is set you get `WorkerConfig.radial_range = [1., inf]`

https://github.com/silx-kit/pyFAI/blob/d00e20c568639d439500d680dfe001275b5406c5/src/pyFAI/io/integration_config.py#L621

The `pyFAI.integrator.azimuthal.AzimuthalIntegrator.integrate1d` gets `radial_range = [1., inf]` and apparently cannot deal with this.

## `pyFAI<2025.1.0`

The worker gets `radial_range` directly from the config dictionary

https://github.com/silx-kit/pyFAI/blob/683a0700eae4fe0b1af8eae9d2e00d946fffda52/src/pyFAI/worker.py#L559-L560

The value of `radial_range` are only set when both `radial_range_min` and `radial_range_max` are not `None`.

## Conclusion

So the issue was not visible in  `pyFAI<2025.1.0`  because `radial_range_min` was simply ignored when `radial_range_max` was not provided.

The question is now: could pyFAI ever handle `radial_range=[1., inf]` or not? If not then `pyFAI.io.integration_config.WorkerConfig` should only have `radial_range is not None` when both min and max are finite.

kif:
close #2586

### DIFF
--- a/src/pyFAI/integrator/azimuthal.py
+++ b/src/pyFAI/integrator/azimuthal.py
@@ -30,7 +30,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "02/09/2025"
+__date__ = "03/09/2025"
 __status__ = "stable"
 __docformat__ = 'restructuredtext'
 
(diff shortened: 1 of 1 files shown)


kif:
@woutdenolf can you have a look, it should at least print out an explicit message and disable the radial/azimuthal semi-defined range, i.e. as previously.

src/pyFAI/integrator/azimuthal.py
```diff
@@ -140,9 +140,17 @@ def integrate1d(self, data, npt, filename=None,
         pos0_scale = unit.scale
 
         if radial_range:
-            radial_range = tuple(radial_range[i] / pos0_scale for i in (0, -1))
+            if numpy.isfinite(radial_range).all():
+                radial_range = tuple(radial_range[i] / pos0_scale for i in (0, -1))
+            else:
+                logger.warning(f"Semi-definied ranges are not supported for radial_range={radial_range}")
```
woutdenolf: `Semi-defined`instead of `Semi-definied`.

woutdenolf: The others as well.

PR #2614, opened 2025-09-26
parallax gui2

kif:
Contains all the other bits & pieces from enabling the parallax correction in calib2 (WIP)


src/pyFAI/detectors/sensors.py
```diff
@@ -153,10 +153,14 @@ def absorbance(self, energy:float, length: float, unit:str="m") -> float:
 Si_MATERIAL = SensorMaterial("Si", 2.329)
 CdTe_MATERIAL = SensorMaterial("CdTe",  5.85)
 GaAs_MATERIAL = SensorMaterial("GaAs", 5.3176)
+Ge_MATERIAL = SensorMaterial("Ge", 5.327)
+Gd2O2S_MATERIAL = SensorMaterial("Gd2O2S", density=7.32)
```
gudlot: @kif Small note, "Gd2O2S" gets here explicitly the density argument. Both are valid, just a comment for consistency

src/pyFAI/gui/dialog/DetectorSelectorDialog.py
```diff
@@ -139,18 +136,71 @@ def __init__(self, parent=None):
...
+            index = self._detectorSensorMaterials.findData(sensor.material)
+            self._detectorSensorMaterials.setCurrentIndex(index)
+            self._detectorSensorParallax.setChecked(True)
+            self._sensorGroup.setEnabled(True)
+        else:
+            self._detectorSensorParallax.setChecked(False)
+            self._sensorGroup.setEnabled(False)
+
+    def __sensorChanged(self, **kwargs):
+        # Finally set the sensor config of all possible the detectors
```
gudlot: `The` is wrong

src/pyFAI/detectors/sensors.py
kif: This is more complicated than it looks: the density of GadOx is correct but we are discussing internally about the actual density of the phosphor which is made of compacted GadOx which value is much less. A reasonnable proper value will be set, with a comment ... maybe this parameter should be user-tuneable.

src/pyFAI/gui/dialog/DetectorSelectorDialog.py
kif: Thanks

PR #2619, opened 2025-09-29
Basic math with IntegrateResult

### LINKED ISSUE #2617: Implement `__add__` and `__sub__` for integrated data objects with proper error propagation
Needed for BM02

kif:
Untested for now, but in theory it shoould work.

close #2617

### DIFF
--- a/src/pyFAI/containers.py
+++ b/src/pyFAI/containers.py
@@ -1409,7 +1409,7 @@ def symmetrize(res2d: Integrate2dResult) -> Integrate2dResult:
 
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
-        intensity = sum_signal / sum_normalization  # noqa: E741
+        intensity = sum_signal / sum_normalization
 
     if res2d.sum_variance is not None:
         sum_variance = _symmetrize_single_array(res2d.sum_variance)


src/pyFAI/integrator/azimuthal.py
```diff
@@ -1185,7 +1185,7 @@ def integrate2d_ng(self, data, npt_rad, npt_azim=360,
             result._set_sum_normalization2(norm2d_sq)
             result._set_sum_variance(var2d)
             result._set_std(std)
-            result._set_std(sem)
+            result._set_sem(sem)
```
loichuder, line 1188: 😄

src/pyFAI/containers.py
```diff
@@ -219,6 +218,83 @@ def __init__(self):
...
+        if reason:
+            raise TypeError(f"Cannot subtract `IntegrateResult` because of {reason}")
+        self._sum_signal -= other.sum_signal
+        if self._sum_variance is None or other.sum_variance is None:
+            self._sum_variance = None
+        else:
+            self._sum_variance += other.sum_variance
+        return self
+
+    def __recalcuate_means__(self):
```
loichuder:
```suggestion
    def __recalculate_means__(self):
```

src/pyFAI/containers.py
```diff
@@ -219,6 +218,83 @@ def __init__(self):
...
+
+    def __add__(self, other):
+        """External add, common part"""
+        reason = self.__are_compatible__(other)
+        if reason:
+            raise TypeError(f"Cannot add `IntegrateResult` because of {reason}")
+        res = copy.deepcopy(self)
+        res._sum_signal = self._sum_signal + other._sum_signal
+        if self._sum_variance is None or other.sum_variance is None:
+            res._sum_variance = None
```
loichuder: If one member of the addition has variance and the other not, should we not keep the variance of the one that has it?

src/pyFAI/containers.py
```diff
@@ -219,6 +218,83 @@ def __init__(self):
...
+        else:
+            res._sum_variance = self._sum_variance + other._sum_variance
```
loichuder, line 246: Math courses are far but the variance of the sum is really the sum of the variances?

src/pyFAI/containers.py
```diff
@@ -1220,7 +1344,7 @@ def rebin1d(res2d: Integrate2dResult) -> Integrate1dResult:
     bins_rad = res2d.radial
     sum_signal = res2d.sum_signal.sum(axis=0)
     sum_normalization = res2d.sum_normalization.sum(axis=0)
-    I = sum_signal / sum_normalization
+    I = sum_signal / sum_normalization  # noqa: E741
```
loichuder: I think the aim of the rule is to encourage something like :


```suggestion
    intensity = sum_signal / sum_normalization
```

src/pyFAI/containers.py
kif: To be discussed ... maybe. Let me think on this.

src/pyFAI/containers.py
kif: Actually, sum-variance is not properly named. Please refer to this paper, 
https://ds.ifi.uni-heidelberg.de/files/Team/eschubert/publications/SSDBM18-covariance-authorcopy.pdf
the `sum_variance` is the quantity called `VWp` in table 1 and the relation to the actual variance is give in eq13.

src/pyFAI/containers.py
kif: Maybe, but what about h,k,l the Miller triplet ?

loichuder: I think only `l` raises the issue but I would say that in this case, it makes sense to keep it and to ignore the rule since it is such a widely-used name for it.

But by avoiding using `I`, we remove any possible confusion with the `l` that we keep :wink:

src/pyFAI/containers.py
loichuder: Not versed enough to understand what is going on in this paper, I'll take your word for it :smile:

PR #2622, opened 2025-10-01
Ensure units can be copied and compared for equality

### LINKED ISSUE #2621: Units cannot be be copied properly
```
In [1]: from pyFAI.units import to_unit

In [3]: q=to_unit("q_nm^-1")

In [4]: q==q
Out[4]: True

In [5]: import copy

In [6]: q==copy.deepcopy(q)
Out[6]: False
```
Blocks #2619

kif:
@loichuder this  PR is needed to enable the basics arithmetics on IntegrationResults you requested.

Do you have an idea why the `id` of all methods changed in the copy  process ? would it be better to patch the way copy works rather than the equality checking ?

Needed for #2619
Close #2621


loichuder:
> This module does not copy types like module, method, stack trace, stack frame, file, socket, window, or any similar types.
> 
> https://docs.python.org/3/library/copy.html

PR #2627, opened 2025-10-03
GUI - parallax (v3)

kif:
work ongoing to define detector sensors in the GUI

### DIFF
--- a/src/pyFAI/gui/tasks/ExperimentTask.py
+++ b/src/pyFAI/gui/tasks/ExperimentTask.py
@@ -25,7 +25,7 @@
 
 __authors__ = ["Valentin Valls", "Jérôme Kieffer"]
 __license__ = "MIT"
-__date__ = "03/11/2025"
+__date__ = "07/11/2025"
 
 import numpy
 import logging
@@ -171,7 +171,7 @@ class ExperimentTask(AbstractCalibrationTask):
         self._energy.setModel(settings.wavelength())
 
         self.__parallaxChangedUpdateModel()  # Initialize & wire
-        self._detectorParallax.clicked.connect(self.__parallaxChangedUpdateModel)
+        self._detectorParallax.toggled.connect(self.__parallaxChangedUpdateModel)
 
         settings.image().changed.connect(self.__imageUpdated)
 
@@ -282,8 +282,8 @@ class ExperimentTask(AbstractCalibrationTask):
             self._detectorSensorName.setText("")
             self._detectorParallax.setChecked(False)
             self._detectorParallax.setVisible(False)
-            self._detectorFileDescription.setVisible(False)
-            self._detectorFileDescriptionTitle.setVisible(False)
+            self._detectorFileDescription.setEnable(False)
+            self._detectorFileDescriptionTitle.setEnabled(False)
 
         else:
             self._detectorLabel.setStyleSheet("QLabel { }")
--- a/src/pyFAI/resources/gui/detector-selection-drop.ui
+++ b/src/pyFAI/resources/gui/detector-selection-drop.ui
@@ -110,7 +110,7 @@ background-color: transparent;
       <item>
        <widget class="QGroupBox" name="groupBox_6">
         <property name="title">
-         <string>Detector orientation: </string>
+         <string>Detector orientation</string>
         </property>
         <layout class="QVBoxLayout" name="verticalLayout_10">
          <item>
--- a/src/pyFAI/resources/gui/calibration-experiment.ui
+++ b/src/pyFAI/resources/gui/calibration-experiment.ui
@@ -324,12 +324,21 @@
            </widget>
           </item>
           <item row="5" column="2">
-           <widget class="QCheckBox" name="_detectorParallax">
+           <widget class="QToolButton" name="_detectorParallax">
+            <property name="sizePolicy">
+             <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+              <horstretch>0</horstretch>
+              <verstretch>0</verstretch>
+             </sizepolicy>
+            </property>
             <property name="toolTip">
-             <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Correct geometry for parrallax effect (much slower)&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+             <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Correct geometry for parrallax effect (much slower).&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
             </property>
             <property name="text">
-             <string/>
+             <string>Plx</string>
+            </property>
+            <property name="checkable">
+             <bool>true</bool>
             </property>
            </widget>
           </item>


kif:
@t20100 I simulated an image at 20keV distance 70mm with parallax effect simulated from a 1mm Si sensor.
Without correction, the system converges to 20.06 keV and 70.44mm distance.
With correction active the system converged to 20.04keV and 69.98mm which is better (even though not perfect).
Without parallax, the results are close to perfect ...

src/pyFAI/gui/dialog/DetectorSelectorDialog.py
```diff
@@ -122,6 +122,7 @@ def __init__(self, parent=None):
         self.__detectorHeight = DataModel()
         self.__pixelWidth = DataModel()
         self.__pixelHeight = DataModel()
+        # self.__sensor = DataModel()
```
kif: can be removed

src/pyFAI/resources/gui/detector-selection-drop.ui
```diff
@@ -157,55 +150,51 @@ background-color: transparent;
          </sizepolicy>
         </property>
         <property name="title">
-         <string/>
+         <string>Sensor:</string>
```
t20100:
```suggestion
         <string>Sensor</string>
```

src/pyFAI/gui/utils/__init__.py
```diff
@@ -83,3 +85,12 @@ def maximize_fig(fig=None):
...
     update_fig(fig)
+
+
+@contextmanager
+def block_signals(w: qt.QWidget):
+    old = w.blockSignals(True)
+    try:
+        yield
+    finally:
+        w.blockSignals(old)
```
t20100: There's one in silx:

https://github.com/silx-kit/silx/blob/731158f028952d6b5d0ed7376053f783daec7381/src/silx/gui/utils/__init__.py#L35

src/pyFAI/gui/tasks/IntegrationTask.py
```diff
@@ -931,7 +933,7 @@ def _initGui(self):
 
         self.__integrationUpToDate = True
         self.__integrationResetZoomPolicy = None
-        method = method_registry.Method(666, "bbox", "histogram", "cython", None)
+        method = method_registry.Method(666, "full", "histogram", "cython", None)
```
t20100, line 936: So `full` becomes the default?

src/pyFAI/gui/tasks/ExperimentTask.py
```diff
@@ -268,8 +280,11 @@ def __detectorModelUpdated(self):
             self._detectorOrientationValue.setText("")
             self._detectorSensorLabel.setText("")
             self._detectorSensorName.setText("")
+            self._detectorParallax.setChecked(False)
+            self._detectorParallax.setVisible(False)
```
t20100, line 284: IMO it's usually best to enable/disable widgets than to show/hide

src/pyFAI/gui/tasks/ExperimentTask.py
```diff
@@ -170,11 +170,15 @@ def _updateModel(self, model):
         self._wavelength.setModel(settings.wavelength())
         self._energy.setModel(settings.wavelength())
 
+        self.__parallaxChangedUpdateModel()  # Initialize & wire
+        self._detectorParallax.clicked.connect(self.__parallaxChangedUpdateModel)
```
t20100: `clicked` is only called upon user interaction and not when the state is changed from the API.
Is it really what you want?
It's usually best to use `toggled` which is emitted for any state change.

src/pyFAI/gui/dialog/DetectorSelectorDialog.py
```diff
@@ -143,50 +143,95 @@ def __init__(self, parent=None):
...
-        self._sensorGroup.setEnabled(self._detectorSensorParallax.isChecked())
-        self.__sensorChanged()
-
-    def getSensorConfig(self):
-        if self._detectorSensorParallax.isChecked():
-            sensor_material = self._detectorSensorMaterials.currentData()
-            sensor = detectors.sensors.SensorConfig(sensor_material)
-            thickness = self._detectorSensorThickness.text()
+        self._detectorSensorThickness.currentIndexChanged.connect(self.__sensorChanged)
+        self._detectorSensorThickness.currentTextChanged.connect(self.__sensorChanged)
```
t20100: Both signals shouldn't be needed:
```suggestion
        self._detectorSensorThickness.currentTextChanged.connect(self.__sensorChanged)
```

src/pyFAI/gui/dialog/DetectorSelectorDialog.py
```diff
@@ -143,50 +143,95 @@ def __init__(self, parent=None):
...
+            if index < 0:
+                self._detectorSensorThickness.addItem(
+                    f"{1e6 * sensor.thickness:4.0f}" if sensor.thickness else "", 
+                    userData=sensor.thickness
+                )
+                index = self._detectorSensorThickness.findData(sensor.thickness)
+            self._detectorSensorThickness.setCurrentIndex(index)
+
+        index = self._detectorSensorMaterials.findData(sensor.material)
+        with block_signals(self._detectorSensorMaterials):
```
t20100, line 226: Is blocking signals really needed here?
Unless I missed something, when calling this method, `__sensorChanged` is not called, so the actual senor info is not updated....
Does it matter if `__sensorChanged` is called twice?

src/pyFAI/gui/dialog/DetectorSelectorDialog.py
```diff
@@ -143,50 +143,95 @@ def __init__(self, parent=None):
...
+                self._detectorSensorMaterials.addItem(" ", userData=None)
+                for key, value in detectors.sensors.ALL_MATERIALS.items():
+                    self._detectorSensorMaterials.addItem(key, userData=value)
+                self._detectorSensorThickness.addItem("", userData=None)
+            self._detectorSensorInfo.setText("|".join(info))
+
+        if isinstance(detector, detectors.Detector):
+            self.setSensorConfig(detector.sensor)
+
+    def getSensorConfig(self) -> SensorConfig:
```
t20100:
```suggestion
    def getSensorConfig(self) -> SensorConfig | None:
```

src/pyFAI/gui/dialog/DetectorSelectorDialog.py
```diff
@@ -143,50 +143,95 @@ def __init__(self, parent=None):
...
-            sensor_material = self._detectorSensorMaterials.currentData()
-            sensor = detectors.sensors.SensorConfig(sensor_material)
-            thickness = self._detectorSensorThickness.text()
+        self._detectorSensorThickness.currentIndexChanged.connect(self.__sensorChanged)
+        self._detectorSensorThickness.currentTextChanged.connect(self.__sensorChanged)
+
+    def _resetSensor(self, detector=None):
+        """populate the 2 comboBox with information from the detector.
+        By default, expose all possible settings"""
+        with block_signals(self._detectorSensorMaterials),  block_signals(self._detectorSensorThickness):
```
t20100, line 166: Is blocking signal really needed?

src/pyFAI/gui/tasks/ExperimentTask.py
kif: Noted.

src/pyFAI/gui/tasks/ExperimentTask.py
kif: Valentin is pushing for replacing this checkbox with a button which could be held-down ... Do you have an opinion on this ?

src/pyFAI/gui/tasks/IntegrationTask.py
kif: For the integration in the calib2-GUI, I think so ... it is in all point superior quality than `BBox` and for this activity, I believe it is correct.
There is an issue (#2662) on persisting the integration method is in the configuration which is even better.

src/pyFAI/gui/dialog/DetectorSelectorDialog.py
kif: I think so: there is one "sensor" with 2 information in it, the "material" and the "thickness".
I block the signals for both of them so that the 2 widgets can get updated synchronously without interfering.

src/pyFAI/resources/gui/detector-selection-drop.ui
kif: Why removing the colon ? Orientation has also one. should it also be removed ?

src/pyFAI/gui/tasks/ExperimentTask.py
kif: Checked in Valentin's code and he is always using `setVisible` in order to remove unused widgets from the display and make it simpler ... I thinks I will stick to his philosophy for calib2...

src/pyFAI/gui/tasks/ExperimentTask.py
kif: I finally implemented QToolButton which is more compact...

src/pyFAI/gui/dialog/DetectorSelectorDialog.py
kif: I checked an `__sensorChanged` was called (too) many times.

src/pyFAI/resources/gui/detector-selection-drop.ui
t20100: > Why removing the colon ? Orientation has also one. should it also be removed ?
Because it's a group box. If orientation is also a group box, there shouldn't be a colon as well.

Other group boxes do not have the colon in the screen shot: https://github.com/silx-kit/pyFAI/pull/2627#issuecomment-3474077386

Issue #2616, opened 2025-09-29
Deprecate python2 style properties ....


kif:
To make this a bit more modern:
replace:
```
def get_smth(self):
    return self._smth

def set_smth(self, value):
    self._smth = value
 smth = property(get_smth, set_smth)
```
with:
```
from pyFAI.utils.decorator import deprecated
[...]
@property
def smth(self):
    return self._smth
@smth.setter
def smth(self, value):
    self._smth = value

# deprecated compatibility layer:
get_smth = deprecated(smth.fget, reason="use property", since_version="2025.09")
set_smth = deprecated(smth.fset, reason="use property", since_version="2025.09")
```

kif:
closed with #2637

PR #2637, opened 2025-10-12
Modernize attributes with @property

gudlot:
Modernizes attributes with `@property` in `src/pyFAI/geometryRefinement.py` and adds a deprecated compatibility layer.  
This is the first step towards replacing old-style getter/setter patterns  
throughout the codebase, as outlined in [silx-kit/pyFAI#2616](https://github.com/silx-kit/pyFAI/issues/2616).

### DIFF
--- a/sandbox/refinment2D.py
+++ b/sandbox/refinment2D.py
@@ -36,7 +36,7 @@ logger = logging.getLogger("pyFAI.refinment2D")
 
 from .azimuthalIntegrator import AzimuthalIntegrator
 from PyMca import SGModule
-
+from pyFAI.utils.decorators import deprecated
 
 class Refinment2D(object):
     """


src/pyFAI/geometryRefinement.py
```diff
@@ -252,7 +306,20 @@ def guess_poni(self, fixed=None):
...
+                            (
+                                cos_tilt
+                                / numpy.sqrt(
+                                    1 - sin_tpr * sin_tpr * sin_tilt * sin_tilt
+                                )
+                            ),
+                        ),
+                    )
+                )  # + or -
                 if cos_tpr * sin_tilt > 0:
```
kif, line 323: sometimes code linters are stupid ... this is a typical case where it is not acceptable.

gudlot: Sorry, overlooked it. Can the linter be modified to avoid this in the future?

kif: Would it be possible to revert the last commit, since it it the one with the linting with  ruff ?

gudlot:
After this last update, I get
```py
Ran 574 tests in 233.934s

FAILED (errors=13, skipped=10)
```
I noticed that the class `Detector` uses still a lot of `splineFile` , but maybe this should be carefully exchanged with `splinefile`.

src/pyFAI/integrator/common.py
```diff
@@ -1658,21 +1658,29 @@ def __save2D(self, filename, intensity, dim1, dim2, error=None, dim1_unit=units.
...
-        self.detector.set_flatfield(flat)
 
-    def get_flatfield(self):
-        return self.detector.get_flatfield()
+    @property
+    def flatfield(self):
+        return self.detector.flatfield
+
+    @flatfield.setter
+    def set_flatfield(self, flat):
```
kif: The pre-commit hook spotted a bug at this place and the name of the method should be `flatfield`, not `set_flatfield`

gudlot: I corrected this mistake. Many thanks for spotting it.

gudlot:
Hi @kif, I noticed a few remaining TODOs in the code that could use a second look. 
- `get_correct_solid_angle_for_spline` in `core.py` -  I'm unsure about the implementation here.  

- The following functions still use the old property methods and I have not yet updated them: 
  - `get_1dsize` in `cli_calibration.py`  
  - `get_name` in `_common.py`  
  - `get_shape` in `refinment2D.py`  

Could you have a look when you get a chance? Thanks!

src/pyFAI/detectors/_common.py
```diff
@@ -1083,60 +1086,71 @@ def set_mask(self, mask):
...
-        if isinstance(value, float):
-            value = value
-        elif isinstance(value, (tuple, list)):
-            value = float(value[0])
-        else:
-            value = float(value)
+    @pixel2.setter
+    def pixel2(self, value):
+        """Set the pixel size along the second dimension."""
+        #TODO: Is this on purpose to take the first entry in tuple, list as pixel2?
```
kif, line 1140: I don't remind the reason for accepting a list/tuple as pixel2 ... could be to set pixel1&pixel2 but it looks pretty ugly.  Maybe one should emit a warning that this a deprecated use-case that will be removed one day. Same for pixel1 probably.

src/pyFAI/geometry/core.py
```diff
@@ -2767,99 +2767,131 @@ def get_shape(self, shape=None):
...
-        if isinstance(value, float):
+        if isinstance(value, float):  # TODO: Is this still necessary?
             self._wavelength = value
         elif isinstance(value, (tuple, list)):
             self._wavelength = float(value[0])
         else:
             self._wavelength = float(value)
         qa = dqa = q_corner = None
         if old_wl and self._wavelength:
+            # TODO: Could use self.qa property here instead of direct _cached_array access?
```
kif: No, because we don't want to trigger the calculation of `qa` if it is not already calculated, but if it calculated, we update it.

The update could be optimized like this:
`qa = (old_wl / self._wavelength) * self._cached_array[....]`
and the same for `q_corner`, it replaces an array division with a float division; thus it saves millions of divisions. 
After those values have been calculates, the cache is emptied and re-populated with the updated values (when available). It is a bit stupid to trash `2th` and `radius` arrays if they exist, they could have been saved and restored as well.

src/pyFAI/geometry/core.py
```diff
@@ -2870,97 +2902,149 @@ def set_wavelength(self, value):
         self.reset()
         # restore updated values
         self._cached_array["q_delta"] = dqa
+        #TODO: This bypasses the qa read only property?
```
kif: Absolutely. You can remove the `?` and the TODO

src/pyFAI/geometry/core.py
```diff
@@ -2990,28 +3078,41 @@ def set_correct_solid_angle_for_spline(self, value):
         v = bool(value)
         with self._sem:
             if v != self._correct_solid_angle_for_spline:
+                # TODO: Confusing - this sets _dssa attribute, not the _dssa property
+                # Should this be clearing the cache via `del self.dssa` instead?
+                # Or is there a historical reason for this attribute assignment?
                 self._dssa = None
```
kif: Well spotted, this should actually remove|set to None any cached_array which key name starts with `solid_angle`. This setter triggers a change of behavior, so all solid_angle arrays should be destroyed.

src/pyFAI/detectors/_common.py
gudlot: Should pixel be just of type `float`? There is a also a similar construction in the `wavelength` method.

src/pyFAI/geometry/core.py
```diff
@@ -2767,99 +2767,131 @@ def get_shape(self, shape=None):
...
+
+    @wavelength.setter
+    def wavelength(self, value):
         "Set the wavelength in meter!"
         old_wl = self._wavelength
-        if isinstance(value, float):
+        if isinstance(value, float):  # TODO: Is this still necessary?
             self._wavelength = value
         elif isinstance(value, (tuple, list)):
             self._wavelength = float(value[0])
```
gudlot, line 2889: According to the documentation, `wavelength` should be of type `float`. 
https://pyfai.readthedocs.io/en/stable/api/geometry/index.html#pyFAI.geometry.core.Geometry.__init__
Is it necessary to keep the check for `tuple`, `list`?

src/pyFAI/geometry/core.py
```diff
@@ -2767,99 +2767,131 @@ def get_shape(self, shape=None):
...
+    def rot3(self):
+        return self._rot3
 
-    def set_rot3(self, value):
-        if isinstance(value, float):
+    @rot3.setter
+    def rot3(self, value):
+        if isinstance(value, float):  # TODO: Is this still necessary?
             self._rot3 = value
         elif isinstance(value, (tuple, list)):
```
gudlot, line 2867: Similar to the `wavelength`, the rotation parameters should be single `float`?

src/pyFAI/geometry/core.py
gudlot: Many thanks for your explanations.

Some more thoughts:
* Would it make sense to add an early return when `old_wl == value` (perhaps with a tolerance of `1e-15`)?
* `self.reset()` clears the entire `self._cached_array` dictionary, correct? As far as I see, only `q_center`, `q_corner`, `q_delta`, and possibly `d*2_center` depend on the wavelength.
* Would it be possible to remove `self.reset()` and update those arrays in place instead?
* Also, what happens to `self._cached_array["q_delta"]`? It seems to be only assigned here.

src/pyFAI/geometry/core.py
gudlot: Are you looking for something like: 
```py
def set_correct_solid_angle_for_spline(self, value):
    v = bool(value)
    with self._sem:
        if v != self._correct_solid_angle_for_spline:
            # Clear ALL solid_angle arrays
            for key in list(self._cached_array.keys()):
                if key.startswith("solid_angle"):
                    self._cached_array[key] = None
            self._correct_solid_angle_for_spline = v
```

src/pyFAI/detectors/_common.py
```diff
@@ -1193,21 +1216,27 @@ def set_flatfiles(self, files, method="mean"):
...
             self.flatfiles = "%s(%s)" % (method, ",".join(files))
 
-    def get_darkcurrent(self):
+    @property
+    def darkcurrent(self):
         return self._darkcurrent
 
-    def get_darkcurrent_crc(self):
+    @property
+    def darkcurrent_crc(self):
```
gudlot: @kif I am unsure if I made here a mistake. Example is now darkcurrent. There are a couple of methods with `_crc` in the name in the code. Should they remain `get_smth_crc` or should they be converted to properties? I converted them to properties, but this will mean change in API.

src/pyFAI/geometry/core.py
kif: * early return, why not, some programmers insists on having a single exit point for any function, but not me. The error should be relative. This would avoid flushing the cache for nothing. After thought, the exact match of the wavelength is preferred, since it could be fit with minor modifications ... Better avoid caching issued there.
* Yes d*2 deserves more math, to be implemented. The issue is that user can define their own units (some do) and one should retrieve the associated unit and check if it depends on lambda in the equation. not that simple.
* This is an option, but then one needs to ensure to reset the engine cache (triggered by reset). Another option is to keep a copy of the dict and populate the new cache with its content (OK for r, 2theta & chi, updated for q).
* The `q_delta` array is also deleted, set to None in practice. It could be recalculated from the former one as well.

Maybe this point deserves a dedicated PR because it can have huge impacts everywhere, especially providing non reproducible results !

src/pyFAI/geometry/core.py
kif: I would deprecate the usage of lists/tuple

src/pyFAI/geometry/core.py
kif: Same: deprecation warning first (for at least a year or two)

src/pyFAI/detectors/_common.py
kif: Those are used to check which arrays have been sent to the GPU. If an array gets reset/recalculated, it could be that it has the same address in memory. The CRC is calculated on all values, not on the memory address.
That said, if there was no property previously, it's better not to change it.

gudlot: The tests ran through, but I will reverse those properties to proper methods. Update: 3 public methods reversed.

src/pyFAI/geometry/core.py
gudlot: Thank you for the feedback. I would also feel more comfortable if this goes into a different PR. I can leave a TODO in the code.

gudlot:
I don't know how this happened, I will correct this in the PR for the deprecation warning.

PR #2626, opened 2025-10-03
Register sensor description in the different detectors

kif:
close #2606

PR #2638, opened 2025-10-12
Extend sensor descriptions in detectors (#2626 follow-up)

gudlot:
### WIP: Extend sensor configuration to detectors

Work in progress for extending sensor configuration to available detectors.  
Builds on [silx-kit/pyFAI#2626](https://github.com/silx-kit/pyFAI/pull/2626).  

- `_rayonix.py`: Mar detectors are missing due to lack of information on sensor configuration.  
- Other detectors: similar knowledge gaps where sensor information is missing.  
- Implemented a warning in the `Detector` constructor when the sensor material is unknown.  
- Unified `SENSORS` as a tuple across touched detectors.  
- Fixed a few typos.  
- Some suggestions are masked as TODO, e.g., `IMMUTABLE_ATTRS` instead of `UNMUTABLE_ATTRS`.

### DIFF
--- a/src/pyFAI/detectors/_dectris.py
+++ b/src/pyFAI/detectors/_dectris.py
@@ -648,7 +648,7 @@ class Pilatus(_Dectris):
         """set the config of the detector
 
         For Pilatus detector, possible keys are: max_shape, module_size, x_offset_file, y_offset_file, orientation, sensor
-        
+
         :param config: dict or JSON serialized dict
         :return: detector instance
         """
--- a/src/pyFAI/detectors/_rayonix.py
+++ b/src/pyFAI/detectors/_rayonix.py
@@ -598,7 +598,7 @@ class Mar345(Detector):
                 raise err
         self.set_pixel1(config.get("pixel1"))
         self.set_pixel2(config.get("pixel2"))
-        
+
         self._orientation = Orientation(config.get("orientation", 3))
         self.sensor = SensorConfig(config["sensor"]) if config.get("sensor") is not None else None
         return self
@@ -624,7 +624,7 @@ class Mar555(Detector):
 
 
     # get_config inherited from Detector; no changes needed
-    
+
     def set_config(self, config):
         """set the config of the detector
 


src/pyFAI/detectors/_common.py
```diff
@@ -118,7 +118,8 @@ class Detector(metaclass=DetectorMeta):
     DUMMY = None
     DELTA_DUMMY = None
     ORIENTATION = 0
-    SENSORS = []
+    SENSORS = ()
+    #TODO: Should this be called IMMUTABLE_ATTRS as in cannot be changed
```
kif: Nop, because IMMUTABLE_ATTRS refers to instance attributes, not class attributes. It is used in the helper function for the copy module.

src/pyFAI/detectors/_common.py
```diff
@@ -260,11 +263,18 @@ def __init__(self,
...
+
+        if isinstance(sensor, dict):
+            sensor = SensorConfig.from_dict(sensor)
+
+        if isinstance(sensor, SensorConfig):
+            if sensor not in self.SENSORS:
+                logger.warning("Sensor %s not in allowed SENSORS: %s", sensor, self.SENSORS)
+            self.sensor = sensor
+        elif sensor is None:
+            logger.warning("No sensor configuration provided; using default behaviour.")
```
kif: This should be at the logging level.

src/pyFAI/detectors/_common.py
```diff
@@ -260,11 +263,18 @@ def __init__(self,
...
+        else:
+            logger.warning("Sensor is of unexpected type: %s", type(sensor))
```
kif, line 277: This one should be at the error level.

src/pyFAI/detectors/_common.py
```diff
@@ -1481,6 +1495,7 @@ def get_config(self):
 
         :return: dict with param for serialization
         """
+        #TODO: Is sensor here missing?
```
kif: Since `sensor` is part of the args of the constructor, it it is missing here.

src/pyFAI/detectors/_rayonix.py
```diff
@@ -582,6 +592,7 @@ def get_config(self):
     def set_config(self, config):
         """set the config of the detector
 
+        #TODO: Copy paste mistake?
```
kif: Why? the returned `dict` should match the constructor's signature.
Indeed, the reference to Eiger is a copy/paste mistake, but the rest of the method looks OK (sensor is still missing).

src/pyFAI/detectors/_common.py
gudlot: I just wanted to point out that in English the correct term is immutable, not unmutable. I understand these are instance attributes that should not be changed.

src/pyFAI/detectors/_common.py
gudlot: Thanks, that makes sense. I’ve adjusted the log levels so that missing sensors are reported with `logger.info`, unregistered ones with `logger.warning`, and unexpected types with `logger.error`.

src/pyFAI/detectors/_rayonix.py
gudlot: Yes, my comment was referring to the word `Eiger`. Should the doc string be extended with `sensor`, `orientation`? 
Re sensor material: I don't have information about this detector. 
I updated the code.

src/pyFAI/detectors/_common.py
kif: OK, sorry for not understanding. This is used only in `detector` and in `geometry.core` in helper function for the copy mechanism of python. The best is to have it addressed in another issue/PR.

src/pyFAI/detectors/_common.py
```diff
@@ -1474,15 +1488,26 @@ def set_config(self, config):
...
         return self
 
     def get_config(self):
         """Return the configuration with arguments to the constructor
 
         :return: dict with param for serialization
         """
-        return {"filename": self._filename,
-                "orientation": self.orientation or 3}
+        #TODO: Or via  super().get_config() ?
```
kif: If it is the same as the parent class, the best is to remove the method.

src/pyFAI/detectors/_dectris.py
```diff
@@ -236,13 +245,16 @@ def set_config(self, config):
         module_size = config.get("module_size")
         if module_size is not None:
             self.module_size = tuple(module_size)
+        #TODO Why is fall back 3 and in init 0?
```
kif: Those orientations are enums. The value 0 describes the absence of configuration, while the value 3 defines the default orientation used in pyFAI. Other value possible (1,2,4) represent flipped images. There are always issues with origin/signs of the azimuthal angles in non-standard representations.

src/pyFAI/detectors/_common.py
gudlot: Thank you. Regarding the other `get_config` methods: could you please clarify the design for when we should rely on the parent implementation versus overriding it in the subclass? I am not fully sure I understand the intended systematic use.

src/pyFAI/detectors/_dectris.py
gudlot: Thank you for the explanation. How should I move forward? Keep it as it is, i.e. `3` ? Should I add `3` as fall back to the other `set_config` methods?

src/pyFAI/detectors/_common.py
gudlot: `NexusDetector.get_config()` is slightly different from `Detector.get_config()` imho. I modified it slightly. I think I understood now better the design of `get_config` and modified it for the detectors I touched.

src/pyFAI/detectors/_dectris.py
gudlot: I kept `3` in the end as fall back.

kif (approved):
LGTM

PR #2655, opened 2025-10-30
Simplification of the detector definitions

kif:
This PR allows to basically remove the constructor method for all classes where the pixel size is enforced.
Tests remain unchanged, so no regression expected.

### DIFF
--- a/src/pyFAI/detectors/_dectris.py
+++ b/src/pyFAI/detectors/_dectris.py
@@ -40,7 +40,7 @@ __status__ = "production"
 import os
 import numpy
 import logging
-from ._common import Detector, Orientation, to_eng, SensorConfig, _ensure_dict
+from ._common import Detector, Orientation, SensorConfig, _ensure_dict
 from ..utils.mathutil import expand2d
 from ..utils.decorators import deprecated_args, deprecated
 logger = logging.getLogger(__name__)


src/pyFAI/detectors/_dectris.py
```diff
@@ -441,9 +442,7 @@ class Mythen(_Dectris):
     force_pixel = True
     MAX_SHAPE = (1, 1280)
     SENSORS = (Si320, Si450, Si1000)
-
-    def __init__(self, pixel1=8e-3, pixel2=50e-6, orientation:int|Orientation=0, sensor:SensorConfig|None=None):
-        super(Mythen, self).__init__(pixel1=pixel1, pixel2=pixel2, orientation=orientation, sensor=sensor)
+    PIXEL_SIZE = (8e-3, 50e-6) 
 
     def get_config(self):
```
gudlot: Could the `get_config` for class `Mythen` be removed? You remove the constructor for `Mythen` and `_Dectris` has no constructor either.

gudlot: In line 460, one could write "Mythen has no masks"

src/pyFAI/detectors/_dectris.py
gudlot, line 1: `Detector.get_config` handles with `self.max_shape`. Can corresponding block in `Pilatus.get_config` be removed, as `Pilatus` inherits from `Detector`?

gudlot: Probably a similar case for `Eiger.get_config`: 
```py
        if ((self.max_shape is not None) and
                ("MAX_SHAPE" in dir(self.__class__)) and
                (tuple(self.max_shape) != tuple(self.__class__.MAX_SHAPE))):
            dico["max_shape"] = self.max_shape
 ```
 Could be removed?

File: src/pyFAI/detectors/_esrf.py
Comment (gudlot) [inline src/pyFAI/detectors/_esrf.py:1]: `Maxipixi.get_config` has this block
```py
if ((self.max_shape is not None) and
                ("MAX_SHAPE" in dir(self.__class__)) and
                (tuple(self.max_shape) != tuple(self.__class__.MAX_SHAPE))):
            dico["max_shape"] = self.max_shape
  ```
  Probably handled by `Detector.get_config` and can be removed?

File: src/pyFAI/detectors/_dectris.py
Comment (gudlot) [reply src/pyFAI/detectors/_dectris.py:1]: Why does `Pilatus4` constructor not have a ` module_size:tuple[int,int]|None=None,` option?

File: src/pyFAI/detectors/_others.py
Comment (gudlot) [inline src/pyFAI/detectors/_others.py:1]: A general comment, also for the other classes in this PR: 
Would it make sense to standardise on super() for consistency and MRO safety?

File: src/pyFAI/detectors/_esrf.py
Comment (gudlot) [inline src/pyFAI/detectors/_esrf.py:1]: Could `Maxipixi. __repr__` be removed? It could use `Detectors` own one.

File: src/pyFAI/detectors/_common.py
Comment (gudlot) [inline src/pyFAI/detectors/_common.py:1]: A suggestion for `NexusDetector.__repr__(self)`

```py
def __repr__(self):
    base = super().__repr__()
    return base.replace(
        f"Detector {self.name}",
        f"{self.name} detector from NeXus file: {self._filename}\t",
        1
    )
```
but I am unsure what should be returned in case of filename is None.

src/pyFAI/detectors/_dectris.py
gudlot: `_Dectris.__repr__(self)` maybe not needed as well?

gudlot (changes requested):
@kif I think I am done with my review.

src/pyFAI/detectors/_dectris.py
```diff
@@ -97,11 +106,32 @@ def get_config(self):
...
+        if "max_shape" in config:
+            self.max_shape = tuple(config["max_shape"])
+        module_size = config.get("module_size")
+        if module_size is not None:
+            self.module_size = tuple(module_size)
+        self._orientation = Orientation(config.get("orientation", 3))
+        self.sensor = SensorConfig(config["sensor"]) if config.get("sensor") is not None else None
+        return self
+
     def __repr__(self):
```
gudlot:
```py
    def __repr__(self):
        """Nice representation of the instance
        """
        txt = f"Detector {self.name}"
        if self.splinefile:
            txt += f"\t Spline= {self.splinefile}"
        if (self._pixel1 is None) or (self._pixel2 is None):
            return "Undefined detector"
        else:
            txt += f"\t PixelSize= {to_eng(self._pixel1)}m, {to_eng(self._pixel2)}m"
        if self.orientation:
            txt += f"\t {self.orientation.name} ({self.orientation.value})"
        if self.sensor:
            txt += f"\t {self.sensor}"
        return txt
```
This method is from the `Detector` class. Would `_Dectris` need its own `__repr__` method?

src/pyFAI/detectors/_common.py
```diff
@@ -1581,3 +1567,16 @@ def getFit2D(self):
...
+
+def _ensure_dict(dico_or_str:str|dict)-> dict:
+    """Helper function decoding a JSON string into a dict if needed"""
+    if isinstance(dico_or_str, dict):
+        config = dico_or_str
+    else:
+        try:
+            config = json.loads(dico_or_str)
+        except Exception as err:  # IGNORE:W0703:
+            logger.error(f"Unable to parse config `{config}` as JSON.\n{type(err).__name__}: {err}")
```
gudlot, line 1578: Would you like to put the `type(err)` in front as in line 192 ? I just noticed the difference.

src/pyFAI/detectors/_dectris.py
gudlot, line 1: Nice ideas. Providing a constructor for `_Dectris` saves code lines.

src/pyFAI/detectors/_esrf.py
```diff
@@ -188,22 +169,10 @@ def set_config(self, config):
...
-        if "max_shape" in config:
-            self.max_shape = tuple(config["max_shape"])
-        module_size = config.get("module_size")
-        if module_size is not None:
-            self.module_size = tuple(module_size)
-        self._orientation = Orientation(config.get("orientation", 3))
-        self.sensor = SensorConfig(config.get("sensor")) if config.get("sensor") is not None else None
+        config = _ensure_dict(config).copy()
+        module_size = config.pop("module_size", None)
+        super().set_config(config)
```
gudlot, line 174: Pretty clean: handling subclass key locally before passing to parent

gudlot (approved):
LGTM. Only two minor comments, e.g. `__repr__` in `_Decris_` imho.

PR #2658, opened 2025-10-31
remove some deprecation warning

kif:
Ready for review ...

### DIFF
--- a/src/pyFAI/gui/peak_picker.py
+++ b/src/pyFAI/gui/peak_picker.py
@@ -459,7 +459,7 @@ class PeakPicker(object):
                 callback(cpt)
         else:
             if not self.points.calibrant.dspacing:
-                logger.error("Calibrant has no line ! check input parameters please, especially the '-c' option")
+                logger.error("Calibrant has no line! Check input parameters please, especially the '-c' option")
                 print(CALIBRANT_FACTORY)
                 raise RuntimeError("Invalid calibrant")
             input("Please press enter when you are happy with your selection" + os.linesep)


src/pyFAI/gui/peak_picker.py
```diff
@@ -458,7 +458,7 @@ def finish(self, filename=None, callback=None):
             if callable(callback):
                 callback(cpt)
         else:
-            if not self.points.calibrant.dSpacing:
+            if not self.points.calibrant.dspacing:
                 logger.error("Calibrant has no line ! check input parameters please, especially the '-c' option")
```
gudlot: No space before ! Capital C for check

src/pyFAI/gui/cli_calibration.py
```diff
@@ -874,8 +868,8 @@ def prompt(self):
                         except ValueError:
                             logger.warning("invalid value")
                         else:
-                            self.geoRef.__getattribute__("set_%s_min" % param)(value_min)
-                            self.geoRef.__getattribute__("set_%s_max" % param)(value_max)
+                            self.geoRef.__getattribute__(f"set_{param}_min")(value_min)
+                            self.geoRef.__getattribute__(f"set_{param}_max")(value_max)
```
gudlot, line 872: Just for my understanding:  This will be later replaced with something like `setattr(obj, "attr_name", value)` , e.g. `setattr(self.geoRef, param + "_max", value_max)`?

kif: Here it could be replaced. But there are other places where it is not possible, because the setter is sent as parameter to a function (especially in calib2) ...

PR #2668, opened 2025-11-05
energy keV <-> wavelength A

### LINKED ISSUE #2667: [calib2] label `wavelength`|`energy` depending on the unit chosen
* "keV" -> energy
* "m" -> wavelength
* "A" -> wavelength

It always annoys me the see a wavelength in keV !

kif:
close #2667

### DIFF
--- a/src/pyFAI/gui/widgets/FitParamView.py
+++ b/src/pyFAI/gui/widgets/FitParamView.py
@@ -25,7 +25,7 @@
 
 __authors__ = ["Valentin Valls", "Jérôme Kieffer"]
 __license__ = "MIT"
-__date__ = "05/11/2025"
+__date__ = "06/11/2025"
 
 from silx.gui import icons
 from silx.gui import qt
@@ -241,7 +241,7 @@ class FitParamView(qt.QObject):
     _iconConstraintMax = None
     _iconConstraintNoMax = None
 
-    def __init__(self, parent, label, internalUnit, displayedUnit=None):
+    def __init__(self, parent, label:str, internalUnit, displayedUnit=None):
         qt.QObject.__init__(self, parent=parent)
         self.__label = label
         self.__labelWidget = qt.QLabel(parent)
@@ -306,8 +306,8 @@ class FitParamView(qt.QObject):
         if self._iconConstraintNoMax is None:
             self._iconConstraintNoMax = icons.getQIcon("pyfai:gui/icons/constraint-no-max")
 
-        displayedUnit.changed.connect(self.changeUnit)
-        self.changeUnit()  # enforcee the relabeling if needed
+        displayedUnit.changed.connect(self.__unitChanged)
+        self.__unitChanged()  # enforce the relabeling if needed, only occures when saved config was with `energy`
 
     def __fireValueAccepted(self):
         self.sigValueAccepted.emit()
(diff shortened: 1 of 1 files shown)


src/pyFAI/gui/widgets/FitParamView.py
```diff
@@ -411,3 +411,14 @@ def __constraintsClicked(self):
...
+    def label(self):
+        """getter for the label"""
+        return self.__label
+
+    def setLabel(self, label):
+        """Change the label"""
+        if label != self.__label:
+            self.__label = label
+            self.__labelWidget.setText(f"{label}:")
+
```
gudlot, line 429: Would it make sense to modernise this to use @property / setter syntax instead of explicit setLabel()? Or is the current form required for Qt compatibility?

src/pyFAI/gui/widgets/FitParamView.py
```diff
@@ -306,6 +306,9 @@ def __init__(self, parent, label, internalUnit, displayedUnit=None):
         if self._iconConstraintNoMax is None:
             self._iconConstraintNoMax = icons.getQIcon("pyfai:gui/icons/constraint-no-max")
 
+        displayedUnit.changed.connect(self.changeUnit)
+        self.changeUnit()  # enforcee the relabeling if needed
```
t20100: Let's keep it private
```suggestion
        displayedUnit.changed.connect(self.__unitChanged)
        self.__unitChanged()  # enforce the relabeling if needed
```

Why passing label to the constructor since it is fully controlled by the displayed unit?

src/pyFAI/gui/widgets/FitParamView.py
```diff
@@ -411,3 +414,20 @@ def __constraintsClicked(self):
 
     def widgets(self):
         return [self.__labelWidget, self.__subLayout, self.__unit, self.__constraints]
+
+    def changeUnit(self):
```
t20100:
```suggestion
    def __unitChange(self):
```

src/pyFAI/gui/widgets/FitParamView.py
```diff
@@ -411,3 +414,20 @@ def __constraintsClicked(self):
...
+        if self.__units[0] == Unit.METER_WL:
```
t20100, line 419: I don't know the logic/structure, but why the fact to update depends on this?

src/pyFAI/gui/widgets/FitParamView.py
```diff
@@ -411,3 +414,20 @@ def __constraintsClicked(self):
...
+            self.setLabel("Energy"
+                        if self.__units[1].value() == Unit.ENERGY
+                        else "Wavelength")
+
+    def label(self):
+        """getter for the label"""
+        return self.__label
```
t20100: Do you need the label outside  of this class?
If no, I'd just remove the getter.

src/pyFAI/gui/widgets/FitParamView.py
```diff
@@ -411,3 +414,20 @@ def __constraintsClicked(self):
...
+
+    def setLabel(self, label):
+        """Change the label, if needed"""
+        if label != self.__label:
+            self.__label = label
+            self.__labelWidget.setText(f"{label}:")
```
t20100, line 428: I would not make this public since the label is updated by the unit

src/pyFAI/gui/widgets/FitParamView.py
kif: We don't want to change the behaviour of the widget when dealing with distances or angles. Just for energy/wavelength.
The `best` way to do this would probably to modify all units (it is an enum) and provide them with an extra label but it was not trivial and I feared many side-effects at other places in the code.

t20100: Then maybe best to have a dedicated widget for energy/wavelength that inherits from this one

PR #2671, opened 2025-11-06
parallax correction instead of displacement:

### LINKED ISSUE #2665: [Parallax] Implement correction from inversion of forward blurring
Right now, it is only the opposite of the blurring which is a coarse approximation.

kif:
close #2665 

Investigations are still ongoing:
* `corrections` are more complicated to calculate than `displacements` since they depend also on the sample detector distance.
* displacement are for the blurring, they are positive.
* corrections are for correction (!), they are negative.
* Numerical values are slightly smaller in absolute value for `correction` (<1%) in comparison to displacements`

After evaluation on synthetic data, the results are always better by 1-10% for corrections. 
Not much, but a bit better.

### DIFF
--- a/src/pyFAI/parallax.py
+++ b/src/pyFAI/parallax.py
@@ -426,7 +426,7 @@ class Parallax:
     due to parallax effect from the sine of the incidence angle
 
     """
-    SIZE = 256  # <8k  best fits into L1 cache
+    SIZE = 1024  # <8k  best fits into L1 cache
 
     def __init__(self, sensor=None, beam=None):
         """Constructor for the Parallax class
@@ -469,9 +469,9 @@ class Parallax:
         if distance != self.distance:
             si = self.sin_incidence
             dr = self.displacement
-            radius = distance * si/(numpy.sqrt(1.0-si**2))  # tan(arcsin())
             with warnings.catch_warnings():
                 warnings.filterwarnings('ignore')
+                radius = distance * si/(numpy.sqrt(1.0-si**2))  # tan(arcsin())
                 self.correction = numpy.interp(radius, radius+dr, radius) - radius  # <0
             self.distance = distance
         return self.correction
@@ -519,6 +519,7 @@ class Parallax:
                 if "class" in bfg:
                     classname = bfg["class"]
                     Klass = globals()[classname]
+                    # Nota: dynamic instanciation could be dangerous, fix this
                 else:
                     Klass = Beam
                 self.beam = Klass()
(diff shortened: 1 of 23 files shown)


src/pyFAI/parallax.py
```diff
@@ -457,16 +460,46 @@ def init(self, over=None):
...
+
+        :param sin_incidence: array with the sine of the incidence angle (can be any dimentionnality)
+        :return: the displacement (bluring, positive)
+        """
         return numpy.interp(sin_incidence, self.sin_incidence, self.displacement)
 
+    def correct(self, sin_incidence:numpy.ndarray, distance:float|None=None) -> numpy.array:
+        """Calculate the correction (in distance, meters) from the sine of the incidence angle
+
+        :param sin_incidence: array with the sine of the incidence angle (can be any dimentionnality)
```
gudlot: It is probably dimensionality

src/pyFAI/parallax.py
```diff
@@ -457,16 +460,46 @@ def init(self, over=None):
...
     def __repr__(self):
         return f"{self.__class__.__name__}({self.sensor!r}, beam={self.beam!r})"
 
     def __str__(self):
         return f"Parallax correction: beam `{self.beam}`, sensor `{self.sensor}`"
 
-    def __call__(self, sin_incidence):
-        """Calculate the displacement from the sine of the incidence angle"""
+    def __call__(self, sin_incidence:numpy.ndarray, distance:float|None=None) -> numpy.array:
+        """Calculate the displacement from the sine of the incidence angle
```
gudlot: Calculate the displacement using the sine of the incidence angle.

src/pyFAI/parallax.py
```diff
@@ -457,16 +460,46 @@ def init(self, over=None):
...
+
+        :param sin_incidence: array with the sine of the incidence angle (can be any dimentionnality)
```
gudlot: dimensionality

src/pyFAI/parallax.py
```diff
@@ -457,16 +460,46 @@ def init(self, over=None):
...
+        :return: the displacement (bluring, positive)
```
gudlot: blurring

src/pyFAI/parallax.py
```diff
@@ -457,16 +460,46 @@ def init(self, over=None):
         self.sin_incidence = numpy.sin(angles)
         self.displacement = numpy.array(displacement)
 
+    def calc_corrections(self, distance:float):
+        """Initialize and cache the correction array
+
+        :param distance: sample-detector distance in meters
+        :return: correction array aas function of sine incidence angle
```
gudlot: as a

PR #2678, opened 2025-11-10
Ensure the splineFile argument is properly read

### LINKED ISSUE #2677: [detector] Cannot be restored with keyword `splineFile` in detector_config
Recent regression ...

kif:
close #2677

### DIFF
--- a/src/pyFAI/detectors/_common.py
+++ b/src/pyFAI/detectors/_common.py
@@ -33,7 +33,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "04/11/2025"
+__date__ = "10/11/2025"
 __status__ = "stable"
 
 import logging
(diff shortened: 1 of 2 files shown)


src/pyFAI/test/test_geometry.py
```diff
@@ -530,6 +530,16 @@ def test_ponifile_custom_detector(self):
         geom.load(ponifile)
         self.assertEqual(geom.detector.get_config(), config)
 
+    def test_ponifile_splineFile(self):
+        "bug #2677"
+        config = {"pixel1": 48.42252e-6, "pixel2": 46.84483e-6, "orientation":3}
+        splinefile = UtilsTest.getimage("frelon.spline")
+        config["splineFile"] = splinefile
```
gudlot, line 537: When do we write "splineFile" with capital F and when with lower f, please? I am a bit confused.

src/pyFAI/detectors/_common.py
```diff
@@ -180,9 +180,10 @@ def factory(cls, name: str, config: Union[None, str, Dict[str, Any]]=None):
         # Create the detector
         detector = None
         if config is not None:
-            config = _ensure_dict(config).copy()
+            config = {key.lower():value for key, value in _ensure_dict(config).items()}  # copy with lower case key
             binning = config.pop("binning", None)
-            kwargs = {key.lower():config.pop(key) for key in inspect.getfullargspec(detectorClass).args if key in config}
+            fullargs_lower = {key.lower() for key in inspect.getfullargspec(detectorClass).args}
+            kwargs = {key:config.pop(key) for key in fullargs_lower if key in config}
```
gudlot, line 186: This is neat. Each detector will now have its config with its own args, dynamically created.

src/pyFAI/test/test_geometry.py
kif: There are thousands of poni-files in the nature and we need to keep pyFAI compatible with them.
The actual culprit file came from ID03 but I built the regression test from files we had already in stock.

I am currently re-running all tutorials and there are other issues related to splineFiles vs splinefile (around Fit2D compatibility).

kif: See #2679

PR #2696, opened 2025-11-20
Validate on python 3.14

kif:
Numba beta version installed to ImageD11


jonwright (approved):

kif:
Well, once it is out, the release will be installed: I hope the 2 lines can be removed.
It requires ImageD11 to pull its dependencies properly... but it also require a release on you side.

PR #2701, opened 2025-11-21
Fix the copy of a `GeometryRefinement` object

### LINKED ISSUE #2602: The copy of `GeometryRefinement` is broken
This is apparently a side effect of the `__copy__()` helper methods defined in `azimuthalIntegrator` or `Geometry`

kif:
close #2602

### DIFF
--- a/src/pyFAI/geometryRefinement.py
+++ b/src/pyFAI/geometryRefinement.py
@@ -42,11 +42,12 @@ import logging
 import numpy
 import math
 from math import pi
+from scipy.optimize import fmin, leastsq, fmin_slsqp
 from .integrator.azimuthal import AzimuthalIntegrator
 from .calibrant import Calibrant, CALIBRANT_FACTORY
 from .utils.ellipse import fit_ellipse
-from .utils.decorators import deprecated
-from scipy.optimize import fmin, leastsq, fmin_slsqp
+from .utils.decorators import deprecated, deprecated_args
+
 
 logger = logging.getLogger(__name__)
 
@@ -90,6 +91,7 @@ class GeometryRefinement(AzimuthalIntegrator):
 
     PARAM_ORDER = ("dist", "poni1", "poni2", "rot1", "rot2", "rot3", "wavelength")
 
+    @deprecated_args({"splinefile":"splineFile"}, since_version="2025.10")
     def __init__(
         self,
         data=None,
@@ -102,7 +104,7 @@ class GeometryRefinement(AzimuthalIntegrator):
         rot3=0,
         pixel1=None,
         pixel2=None,
-        splineFile=None,
+        splinefile=None,
         detector=None,
         wavelength=None,
         **kwargs,
@@ -122,7 +124,7 @@ class GeometryRefinement(AzimuthalIntegrator):
         :param rot3: guessed tilt of the detector around the incoming beam axis (optional, in rad)
         :param pixel1: Pixel size along the vertical direction of the detector (in m), almost mandatory
         :param pixel2: Pixel size along the horizontal direction of the detector (in m), almost mandatory
-        :param splineFile: file describing the detector as 2 cubic splines. Replaces pixel1 & pixel2
+        :param splinefile: file describing the detector as 2 cubic splines. Replaces pixel1 & pixel2
         :param detector: name of the detector or Detector instance. Replaces splineFile, pixel1 & pixel2
         :param wavelength: wavelength in m (1.54e-10)
 
@@ -146,7 +148,7 @@ class GeometryRefinement(AzimuthalIntegrator):
         if (
             (pixel1 is None)
             and (pixel2 is None)
-            and (splineFile is None)
+            and (splinefile is None)
             and (detector is None)
         ):
             raise RuntimeError(
@@ -161,7 +163,7 @@ class GeometryRefinement(AzimuthalIntegrator):
             rot3,
             pixel1,
             pixel2,
-            splineFile,
+            splinefile,
             detector,
             wavelength=wavelength,
             **kwargs,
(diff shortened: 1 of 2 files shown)


src/pyFAI/geometryRefinement.py
```diff
@@ -71,6 +71,23 @@
 
 
 class GeometryRefinement(AzimuthalIntegrator):
+    _IMMUTABLE_ATTRS = AzimuthalIntegrator._IMMUTABLE_ATTRS + (
```
gudlot, line 75: Probably all float in the tuple.
Is it given that the attributes exist always?

src/pyFAI/geometryRefinement.py
```diff
@@ -189,9 +206,23 @@ def __init__(
         self._wavelength_min = 1e-15
         self._wavelength_max = 100.0e-10
 
+    def __copy__(self):
+        """:return: a shallow copy of itself."""
+        new = self.__class__(data=self.data,
+                             detector=self.detector,
+                             calibrant = self.calibrant)
+        for key in self._IMMUTABLE_ATTRS:
+            new.__setattr__(key, self.__getattribute__(key))
```
gudlot, line 217: This line ` new.__setattr__(key, self.__getattribute__(key))` requires that attributes exist, for example in the constructor

src/pyFAI/geometryRefinement.py
```diff
@@ -189,9 +206,23 @@ def __init__(
...
+        new.param = [new._dist, new._poni1, new._poni2, new._rot1, new._rot2, new._rot3]
```
gudlot: Wavelength is not required because it is kept constant?

src/pyFAI/geometryRefinement.py
gudlot, line 1: There is in `__deepcopy__` written `splineFile=splineFile`. I am unsure about the writing of `splineFile` here.
Also line 204: self.detector.splineFile

There is later 
 `if "copy" in dir(old_value):`.
 I think it could be replaced with 
 ```py
 if hasattr(old_value, "copy"):
    new_value = old_value.copy()
 ```   

@kif : Thank you for showing me `__deepcopy__`. It avoids redundant copying by using the memo dictionary, so shared references and cycles are handled safely.

src/pyFAI/test/test_geometry_refinement.py
```diff
@@ -839,6 +840,8 @@ def test_synthetic(self):
         mycalibrant.wavelength = 1e-10
         r2 = GeometryRefinement(data, calibrant=mycalibrant, detector="Fairchild",
                                 wavelength=mycalibrant.wavelength)
+        r3 = copy.copy(r2)
+        r4 = copy.copy(r2)
```
gudlot: Do you want r4 to be a deepcopy? Then it would be `r4 = copy.deepcopy(r2)`

src/pyFAI/test/test_geometry_refinement.py
```diff
@@ -855,6 +858,24 @@ def test_synthetic(self):
             self.assertAlmostEqual(ref[key][0], r2.__getattribute__(key), delta=ref[key][1],
                                    msg="%s is %s, I expected %s%s%s" % (key, r2.__getattribute__(key), ref[key], os.linesep, r2))
 
+        # test the copy
+        self.assertEqual(r3.calibrant, r2.calibrant)
```
gudlot, line 862: Are there more immutable internals or attributes to assert?

src/pyFAI/test/test_geometry_refinement.py
```diff
@@ -855,6 +858,24 @@ def test_synthetic(self):
...
+        self.assertTrue(numpy.all(r3.data==r2.data))
+        r3.guess_poni()
+        r3.refine2(10000000, fix=[])
+        for key in ref.keys():
+            self.assertAlmostEqual(r3.__getattribute__(key), r2.__getattribute__(key), delta=ref[key][1],
+                                   msg="%s is %s, I expected %s%s%s" % (key, r3.__getattribute__(key), ref[key], os.linesep, r3))
+
+        # test the deep-copy
+        self.assertEqual(r4.calibrant, r2.calibrant)
+        self.assertTrue(numpy.all(r4.data == r2.data))
```
gudlot, line 874: One could add a small mutation test? `r4.data[...]= something` and assert `r2.data[...]` remains unchanged. 

`np.testing.assert_array_equal` for comparing arrays?

src/pyFAI/test/test_geometry_refinement.py
```diff
@@ -855,6 +858,24 @@ def test_synthetic(self):
...
+        r4.guess_poni()
+        r4.refine2(10000000, fix=[])
+        for key in ref.keys():
+            self.assertAlmostEqual(r4.__getattribute__(key), r2.__getattribute__(key), delta=ref[key][1],
+                                   msg="%s is %s, I expected %s%s%s" % (key, r4.__getattribute__(key), ref[key], os.linesep, r4))
+
```
gudlot, line 892: Maybe a test for `new._cached_array `?

src/pyFAI/test/test_geometry_refinement.py
gudlot, line 1: Maybe it’s worth considering tests that explicitly check object identity using `assertIs(a, b)` for shared/immutable attributes and `assertIsNot(a, b)` for mutable attributes that should be independent.

src/pyFAI/geometryRefinement.py
kif: Those are all declared and are immutable objects, so can be simply copied, no extra care needed.

src/pyFAI/geometryRefinement.py
kif: They should indeed all be declared in constructor ...

src/pyFAI/geometryRefinement.py
kif: Could be improved, indeed.

src/pyFAI/test/test_geometry_refinement.py
kif: Not sure, it would constrain the object to be exactly the same, and mutability often bits. For small objects, it is simpler to check equality.

PR #2705, opened 2025-11-25
Fuji Imaging-Plate data

kif:
Needed for #2699

### DIFF
--- a/src/pyFAI/test/test_detector.py
+++ b/src/pyFAI/test/test_detector.py
@@ -433,7 +433,7 @@ class TestDetector(unittest.TestCase):
         self.assertTrue(fs2==fs)
     
     def test_sensor3(self):
-        self.assertAlmostEqual(sensors.BaFBr085I015_MATERIAL.absorbance(20, 1e-4), 0.6538, 4)
+        self.assertAlmostEqual(sensors.BaFBr085I015_MATERIAL.absorbance(20, 1e-4), 0.6538, 3)
 
 
 


src/pyFAI/detectors/sensors.py
```diff
@@ -156,7 +156,7 @@ def absorbance(self, energy:float, length: float, unit:str="m") -> float:
 ALL_MATERIALS["CdTe"] = CdTe_MATERIAL = SensorMaterial("CdTe", density=5.85)
 ALL_MATERIALS["GaAs"] = GaAs_MATERIAL = SensorMaterial("GaAs", density=5.3176)
 ALL_MATERIALS["Gd2O2S"] = Gd2O2S_MATERIAL = SensorMaterial("Gd2O2S", density=7.32)
-
+ALL_MATERIALS["BaFBr0.85I0.15"] = BaFBr085I015_MATERIAL =SensorMaterial("BaFBr0.85I0.15", density=3.3)
```
gudlot, line 160: According to https://theses.hal.science/tel-04521313v1/file/SONG_PANYI_2023.pdf, p94, the density is 3.18g/cm^3, at least that is my understanding.
I would like to suggest to add as comment : `#https://theses.hal.science/tel-04521313v1/file/SONG_PANYI_2023.pdf, p94` as source.
I will change in my branch the key towards `BaFBr0.85I0.15`

gudlot (commented):
Which density is the correct one for `BaFBr085I015`?

src/pyFAI/detectors/sensors.py
kif: Crystallographic density is often higher than the measured ones for those phosphor due to the difficulty to manufacture compact material. The figure 3.18 is probably more correct.

gudlot: Is the consent then, 3.18 will be used? In that case `SensorMaterial("BaFBr0.85I0.15", density=3.18)`

PR #2715, opened 2025-12-05
[pyFAI-integrate] Complete fiber API

EdgarGF93:


EdgarGF93:
Which Integrator save when both contain parameters? -> The active tab at the moment of clicking save config

EdgarGF93:
The GUI is growing...
Screenshot from 2025-12-05 19-25-22

EdgarGF93:
I think I'm done with the plumbing between GUI, workerconfig and worker

Issue #2769, opened 2026-01-27
[FiberIntegrator] Update some docs inherited from AzimuthalIntegrator


EdgarGF93:

EdgarGF93:
Closed with #2772

PR #2772, opened 2026-01-27
[Fibert-Unit]simplify code, fix numpy equations, shortcuts for qlab

EdgarGF93:


EdgarGF93:
It addresses #2773, #2771 and #2769

doc/source/usage/tutorial/FiberGrazingIncidence.ipynb
kif, line 1: Looks like this notebook was replayed with an elder version of pyFAI ...

PR #2776, opened 2026-01-30
Azimuthal variance accumulation

kif:
Propagate "properly" also azimuthal variance.

It makes sense to do it with `union` but I doubt it can be properly calculated with `add` or `subtract` ...

Needs to be checked with a demonstrator ... probably an extension to the Spotty tutorial

### DIFF
--- a/src/pyFAI/containers.py
+++ b/src/pyFAI/containers.py
@@ -30,7 +30,7 @@ __authors__ = ["Valentin Valls", "Jérôme Kieffer"]
 __contact__ = "valentin.valls@esrf.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "30/01/2026"
+__date__ = "27/02/2026"
 __status__ = "development"
 
 import math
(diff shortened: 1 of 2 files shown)


src/pyFAI/containers.py
```diff
@@ -338,6 +341,16 @@ def union(self, other):
         else:
             res._sum_variance = self._sum_variance + other._sum_variance
             res._sum_normalization2 = self._sum_normalization2 + other._sum_normalization2
+            if self.error_model == ErrorModel.AZIMUTHAL:
+                # Crossed term:
+                delta = (self._sum_signal * other._sum_normalization - other._sum_signal*self._sum_normalization)
+                if numpy.nansum(other._sum_normalization2) < numpy.nansum(self._sum_normalization2):
+                    tmp = other._sum_normalization2 / other._sum_normalization
```
t20100: Why deviating from how eq 18 is written in https://journals.iucr.org/j/issues/2025/01/00/jo5109/index.html?

Image

Maybe choose a better name than `tmp` to help the readers

src/pyFAI/containers.py
```diff
@@ -338,6 +341,16 @@ def union(self, other):
         else:
             res._sum_variance = self._sum_variance + other._sum_variance
             res._sum_normalization2 = self._sum_normalization2 + other._sum_normalization2
+            if self.error_model == ErrorModel.AZIMUTHAL:
+                # Crossed term:
+                delta = (self._sum_signal * other._sum_normalization - other._sum_signal*self._sum_normalization)
+                if numpy.nansum(other._sum_normalization2) < numpy.nansum(self._sum_normalization2):
```
t20100: Is this for numerical stability? other reason?

t20100 (approved):

src/pyFAI/containers.py
kif: The formula is not "exact" but closer to when the A is larger than B. 
Σnormalization² is positive, this is why  we use it to assess the size of the ensemble.

src/pyFAI/containers.py
kif: The implementation based on the formula is not always the most efficient...
Agreed for `tmp`

PR #2784, opened 2026-02-03
Fix pilx for empty mask

### LINKED ISSUE #2782: Diffmap with generic detector + mask
Generic detector (150µm pixel, ... to simulate a binned Eiger2 9M)
Mask saved as MSK file ...
Really strange bug since the mask is apparently loaded as list of list containing 1 and None ... not as a numpy array.

Processing OK but when loading the result in pilx fails.

kif:
close #2782


loichuder (approved):
👍

kif:
Thanks Loic.

PR #2786, opened 2026-02-03
move methods from azimuthal to common

EdgarGF93:


kif (changes requested):
I agree with the modification made but most new methods/function lack proper doc-strings. Typing does not replace actual documentation.

PR #2814, opened 2026-03-06
Create AGENTS.md

kif:


AGENTS.md
```diff
@@ -0,0 +1,25 @@
...
+One cannot simply append `src` to the PYTHONPATH.
+For simple tries, it is possible to run:
+`python bootstrap.py` to launch a python interpreter with the PATH properly setup
+
+## Testing
+Tests can be run with `python ./run-tests.py` which will compile a local version of extensions, and run all tests within this local installation.
+
+## Docstrings
+All functions/classes require a proper docstring. 
+Typing is a bonus but not mandatory.
```
gudlot: How about this? In addition, this might be a good place to define a single docstring convention for pyFAI.
Would it make sense to adopt the NumPy style?

````
## Docstrings
All functions/classes require a proper docstring.
The docstring must describe the purpose, inputs, and outputs of the function in human-readable form.
Use NumPy-style docstrings if possible.
Typing hints are a bonus but not mandatory.
````

File: AGENTS.md [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+
+## Role
+You are a senior developer. 
+Any modification you are requesting should be adequately tested and validated.
```
gudlot: Suggestion
```
You are a senior scientific software developer.  
Any modification you propose must be adequately tested and validated using pyFAI's testing framework.  
Implementation details should go in comments, not in docstrings.
```

gudlot (commented):
A chance for unified docstrings?

kif:
There is a comment on the usage of AI agents in #2824 (which was apparently done with them). This reactivates the interest for the topic.

gudlot:
> There is a comment on the usage of AI agents in #2824 (which was apparently done with them). This reactivates the interest for the topic.

The comment says "This could enable Natural Language Processing (NLP) of Synchrotron/X-ray diffraction data, making the ecosystem highly accessible for users with less background in diffraction physics."

To the best of my knowledge, NLP deals with semantics rather than diffraction data itself. The actual application here is instructing an agent to perform pyFAI tasks. For this to work correctly, a model would need to be trained to understand pyFAI instructions through correct prompt engineering and a high-quality model. I assume the existing pyFAI tests could be a valuable resource to help guide and validate such a model. A RAG, maybe knowledge based graph, would be the direction to go.

kif:
Integrated as part of 2026.09

PR #2816, opened 2026-03-06
Fix serveral minor issues on windows

### LINKED ISSUE #2815: [windows] applications do not acccept options ...
Example:
`pyFAI-integrate --help` does not provide options

Solution foreseen:
* [ ] Add entry-point console-script in the `pyproject.toml` for `pyFAI-integrate-cli`, but only for windows.
* [ ] Update the documentation accordingly

kif:
close #2815

### DIFF
--- a/doc/source/operations/windows.rst
+++ b/doc/source/operations/windows.rst
@@ -68,21 +68,6 @@ To upgrade the C/C++-code in pyFAI, one needs in addition Cython:
    pip install -r requirements.txt --upgrade
    pip install . --upgrade
 
-Idiosyncrasy
-------------
-
-In order to be able to run and debug pyFAI-related application on windows' terminal (cmd.exe), each GUI application exists also with the -cli suffix:
-* pyFAI-calib2 -->  pyFAI-calib2-cli
-* pyFAI-integrate -->  pyFAI-integrate-cli
-* pyFAI-diffmap -->  pyFAI-diffmap-cli
-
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -43,6 +43,35 @@ endif
...
+  acceptable = ['msvc', 'clang-cl']
+  warning_msg = '''
+    -----------------------------------------------------------------
+    Detected @0@ compiler on Windows: @1@
+    This project is primarily tested with Microsoft Visual C++.
+    If you encounter problems, switch to MSVC or add the needed
+    patches for your compiler.
+    -----------------------------------------------------------------
+  '''
+  # C compiler check
```
gudlot, line 61: A suggestion: You could introduce a list like `unsupported=[]`. If `c_id` and `cpp_id` are not in `acceptable`, they fill up `unsupported`. 
If `unsupported` has a length of larger than 0, then `meson.error`  happens, maybe stating what kind of non-supported compiler was found and which are allowed.
Does one need the warnings?

gudlot (commented):
Add a list `unsopported` and have one fatal error only?

meson.build
kif: For now, I am not sure it is the right solution ... when `msys` is installed on windows computer it often results in non-usable binaries or missing libraries. The same could occur under macos with gcc installed beside apple's compiler (clang-like)

PR #2821, opened 2026-03-19
Implement renormalization of diffraction data

### LINKED ISSUE #2737: Renormalize IntegrationResult ?
Check it this would be possible without re-accessing the input image ?

kif:
* Implement renormalization of integration results (close #2737)
* Fix regression in recalculate_means
* test against the regression spotted

Nota: Switch to NumExpr since the math are done in parallel and formula have a better readability.

### DIFF
--- a/src/pyFAI/containers.py
+++ b/src/pyFAI/containers.py
@@ -334,8 +334,8 @@ class IntegrateResult(_CopyableTuple):
         """
         other = deepcopy(self) if copy else self
         ratio = value/self.normalization_factor
-        other.sum_normalization *= ratio
-        other.sum_normalization2 *= ratio**2
+        other._sum_normalization *= ratio
+        other._sum_normalization2 *= ratio**2
         other._set_normalization_factor(value)
         return other.__recalculate_means__()
 


src/pyFAI/ext/splitPixel.pyx
```diff
@@ -485,8 +485,8 @@ def fullSplit1D_engine(pos not None,
                           numpy.asarray(out_data[:, 1]) if error_model else None,
                           numpy.asarray(out_data[:, 2]),
                           numpy.asarray(out_data[:, 3]),
```
kif, line 487: This was a bug, spotted accidentally while investigating another one

src/pyFAI/test/test_containers.py
```diff
@@ -75,6 +75,21 @@ def setUpClass(cls):
     def tearDownClass(cls):
         cls.img = cls.ai = None
```
kif, line 77: This is the non regression test because `recalculate_means` was giving nonsense answers

src/pyFAI/test/test_containers.py
```diff
@@ -288,6 +303,16 @@ def test_maths(self):
         # self.assertTrue(numpy.allclose(c.sem, b.sem, rtol=0.4))
         # self.assertTrue(numpy.allclose(c.sigma, b.sigma, rtol=0.4))
```
kif, line 305: This is the non regression test for renormalization, the actual purpose of this PR

src/pyFAI/containers.py
```diff
@@ -303,24 +304,41 @@ def __isub__(self, other):
         return self.__recalculate_means__()
```
kif, line 303: Re-implementation of `recalculate_means` which should be correct now. 
--> Uses NumExpr

src/pyFAI/containers.py
```diff
@@ -41,11 +41,12 @@
 from enum import IntEnum
 from dataclasses import fields, asdict  # noqa
 import numpy
+import numexpr
 from numpy.typing import ArrayLike
 from .utils.dataclasses import dataclass, case_insensitive_dataclass  # noqa
 from .utils.decorators import deprecated_warning
 
-
+copy_module = copy
```
t20100:
```suggestion
from copy import deepcopy
```
or 
```suggestion
import copy as _copy
```
maybe

src/pyFAI/containers.py
```diff
@@ -303,24 +304,41 @@ def __isub__(self, other):
...
+    def renormalize(self, value:float, copy=True):
+        """Recalculate the diffraction pattern with a different normalization factor
+
+        :param value: new normalization factor
+        :param copy: leave the current object untouched if True, else mangle-it in place
+        :return: IntegrateResult instance
+        """
+        other = copy_module.deepcopy(self) if copy else self
+        other.sum_normalization[...] *= value/self.normalization_factor
+        other.sum_normalization2[...] *= value**2/self.normalization_factor**2
```
t20100:
```suggestion
        other.sum_normalization *= value/self.normalization_factor
        other.sum_normalization2 *= value**2/self.normalization_factor**2
```

src/pyFAI/containers.py
```diff
@@ -303,24 +304,41 @@ def __isub__(self, other):
...
-                        out=self.intensity, where=nomask)
-        self.intensity[mask] = self.dummy
-        if self.sum_variance is not None:
-            numpy.sqrt(self._sum_variance, out=self._sem)
-            numpy.divide(self._sem, self.sum_normalization,
-                         out=self._sem, where=nomask)
-            self._sem[mask] = self.dummy
-
-            self._std = numpy.divide(self._sum_variance, self._sum_normalization2,
-                                    out=self._sem, where=nomask)
```
t20100, line 318: So the error was `out=self._sem` here.

kif: indeed ... it is possible to make numpy fast but there are always tons of array allocation. The NumExpr solution is cleaner .

PR #2833, opened 2026-03-27
`getSaveFileName` call compatible with PySide6

### LINKED ISSUE #2832: [diffmap] PySide6 vs PyQt6 behavour change
See images, on windows. Works on PyQt6, fails on PySide6

Image
Image

kif:
close #2832

### DIFF
--- a/src/pyFAI/gui/diffmap_widget.py
+++ b/src/pyFAI/gui/diffmap_widget.py
@@ -528,8 +528,8 @@ class DiffMapWidget(qt.QWidget):
 
     def save_config(self):
         logger.info("DiffmapWidget.save_config()")
-        json_file = qt.QFileDialog.getSaveFileName(caption="Save configuration as json",
-                                                   directory=self.json_file,
+        json_file = qt.QFileDialog.getSaveFileName("Save configuration as json",
+                                                   self.json_file,
                                                    filter="Config (*.json)")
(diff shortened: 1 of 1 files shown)


src/pyFAI/gui/diffmap_widget.py
```diff
@@ -528,8 +528,8 @@ def restore(self, fname=None):
 
     def save_config(self):
         logger.info("DiffmapWidget.save_config()")
-        json_file = qt.QFileDialog.getSaveFileName(caption="Save configuration as json",
-                                                   directory=self.json_file,
+        json_file = qt.QFileDialog.getSaveFileName("Save configuration as json",
+                                                   self.json_file,
```
t20100: According to https://doc.qt.io/qt-6/qfiledialog.html#getSaveFileName, the first argument should be the parent which is missing here, so I would expect this to fails
Otherwise, passing the first 3 arguments as positional arguments looke to be the only way to be PyQt6/PySide6 compatible.


```suggestion
        json_file = qt.QFileDialog.getSaveFileName(None, # Or the widget that should be the parent
                                                  "Save configuration as json",
                                                   self.json_file,
```

kif: Indeed. Thanks for spotting it.

PR #2840, opened 2026-04-08
calib2 CLI parsing

### LINKED ISSUE #2839: [calib2] `--poni` option does not populate the wavelength in the Experiment settings
The PONI-file is properly read and the detector is properly setup.

kif:
Fix parsing of:
* wavelength
* ponifile
* dark
* flat

close #2839

### DIFF
--- a/src/pyFAI/app/calib2.py
+++ b/src/pyFAI/app/calib2.py
@@ -217,12 +217,11 @@ def configure_parser_arguments(parser):
                         " q_A^-1, r_mm. Default: 2th_deg", type=str, default="2th_deg")
     # Not yet used
     parser.add_argument("--no-gui", dest="gui",
-                        help="Force the program to run without a Graphical interface",
+                        help="Run without a GUI",
                         default=True, action="store_false")
     # Not yet used
     parser.add_argument("--no-interactive", dest="interactive",
-                        help="Force the program to run and exit without prompting"
-                        " for refinements", default=True, action="store_false")
+                        help="Run and exit without prompting. Unused", default=True, action="store_false")
 
     # From Calibration
     # Not yet used
--- a/src/pyFAI/app/calib2.py
+++ b/src/pyFAI/app/calib2.py
@@ -28,7 +28,7 @@ __author__ = "Valentin Valls"
 __contact__ = "valentin.valls@esrf.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "09/04/2026"
+__date__ = "10/04/2026"
 __status__ = "production"
 
 import os
@@ -399,6 +399,10 @@ def setup_model(model, options):
                 logger.warning("Calibrant enforced in the command line. Calibrant from --poni argument ignored.")
                 poniFile.extra["Calibrant"] = local_calibrant
 
+            if len(args) == 0 and poniFile.extra.get("Image"):
+                logger.warning("No input files specified. Using poni-file's calibration image.")
+                args = [poniFile.extra.get("Image")]
+
             geometryFromPoni.wavelength().setValue(wavelength)
             geometryHistory = model.geometryHistoryModel()
             now = datetime.datetime.now()
@@ -406,8 +410,10 @@ def setup_model(model, options):
         else:
             logger.warning("PONI file '%s' do not exists. --poni option ignored.", options.poni)
 
-    settings.detectorModel().setDetector(detector)
-    settings.wavelength().setValue(wavelength)
+    if detector:
+        settings.detectorModel().setDetector(detector)
+    if wavelength:
+        settings.wavelength().setValue(wavelength)
     if local_calibrant:
         try:
             if local_calibrant in calibrant.CALIBRANT_FACTORY:
--- a/src/pyFAI/app/calib2.py
+++ b/src/pyFAI/app/calib2.py
@@ -28,7 +28,7 @@ __author__ = "Valentin Valls"
 __contact__ = "valentin.valls@esrf.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "05/12/2025"
+__date__ = "08/04/2026"
 __status__ = "production"
 
 import os
@@ -326,9 +326,8 @@ def setup_model(model, options):
     args = options.args
 
     # The module must not import the GUI
-    from pyFAI.gui.utils import units
-    # TODO: This should be removed
-    import pyFAI.gui.cli_calibration
+    from ..gui.utils import units as gui_units
+    from ..gui.model.GeometryModel import GeometryModel
 
     # Settings
     settings = model.experimentSettingsModel()
@@ -347,20 +346,19 @@ def setup_model(model, options):
         if local_calibrant:
             settings.calibrantModel().setCalibrant(local_calibrant)
 
+    wavelength = None
     if options.wavelength:
-        value = units.convert(options.wavelength, units.Unit.ANGSTROM, units.Unit.METER_WL)
-        settings.wavelength().setValue(value)
+        wavelength = gui_units.convert(options.wavelength, gui_units.Unit.ANGSTROM, gui_units.Unit.METER_WL)
 
     if options.energy:
-        value = units.convert(options.energy, units.Unit.ENERGY, units.Unit.METER_WL)
-        settings.wavelength().setValue(value)
+        wavelength = gui_units.convert(options.energy, gui_units.Unit.ENERGY, gui_units.Unit.METER_WL)
 
     if options.polarization_factor:
         settings.polarizationFactor().setValue(options.polarization_factor)
 
     if options.detector_name:
         try:
-            detector = pyFAI.gui.cli_calibration.get_detector(options.detector_name)
+            detector = detectors.detector_factory(options.detector_name)
             if options.pixel:
                 logger.warning("Detector model already specified. Pixel size argument ignored.")
         except Exception as e:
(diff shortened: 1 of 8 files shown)


src/pyFAI/app/calib2.py
```diff
@@ -207,24 +204,24 @@ def configure_parser_arguments(parser):
...
     # Not yet used
     parser.add_argument("--saturation", dest="saturation",
-                        help="consider all pixel>max*(1-saturation) as saturated and "
+                        help="Consider all pixel>max*(1-saturation) as saturated and "
                         "reconstruct them, default: 0 (deactivated)",
                         default=0, type=float)
     # Not yet used
     parser.add_argument("--weighted", dest="weighted",
-                        help="weight fit by intensity, by default not.",
+                        help="Weight the fit by intensity of points, by default not.",
```
dk-teknologisk-gnlo: Suggestion: `Use point intensity as weights for fitting. Default: off.`

src/pyFAI/app/calib2.py
```diff
@@ -207,24 +204,24 @@ def configure_parser_arguments(parser):
...
                         default=False, action="store_true")
     # Not yet used
     parser.add_argument("--unit", dest="unit",
                         help="Valid units for radial range: 2th_deg, 2th_rad, q_nm^-1,"
                         " q_A^-1, r_mm. Default: 2th_deg", type=str, default="2th_deg")
     # Not yet used
     parser.add_argument("--no-gui", dest="gui",
-                        help="force the program to run without a Graphical interface",
+                        help="Force the program to run without a Graphical interface",
```
dk-teknologisk-gnlo: Suggestion: `Run without a GUI`

src/pyFAI/app/calib2.py
```diff
@@ -207,24 +204,24 @@ def configure_parser_arguments(parser):
...
                         default=True, action="store_false")
     # Not yet used
     parser.add_argument("--no-interactive", dest="interactive",
-                        help="force the program to run and exit without prompting"
+                        help="Force the program to run and exit without prompting"
```
dk-teknologisk-gnlo: Suggestion: `Run and exit without prompting.`

src/pyFAI/app/calib2.py
```diff
@@ -389,30 +383,40 @@ def setup_model(model, options):
...
-
-                geometryHistory = model.geometryHistoryModel()
-                now = datetime.datetime.now()
-                geometryHistory.appendGeometry("Ponifile", now, geometryFromPoni, None)
+            if wavelength is None:
+                wavelength = poniFile.wavelength
             else:
-                logger.warning("Detector redefined in the command line. Detector from --poni argument ignored.")
+                logger.warning("Wavelength/Energy enforced in the command line. Wavelength from --poni argument ignored.")
+                poniFile._wavelength = wavelength
```
dk-teknologisk-gnlo, line 394: I currently don't understand, why this attribute has an `_`.  Line 401: No underscore. Or is it here correct and in line 401 not correct? See
https://github.com/silx-kit/pyFAI/blob/19b40a44cfcf97db8c91de5291b03aea0d999c66/src/pyFAI/io/ponifile.py#L70

dk-teknologisk-gnlo (commented):
@kif I had a look and left some comments. Looking forward to hearing what you think about them.

src/pyFAI/app/calib2.py
kif: thanks

kif:
GPT5 review:
Summary / verdict
- I fetched and inspected PR #2840 (calib2 CLI parsing) and the 7 modified files.
- Overall the changes look well scoped: improved CLI help text, more robust handling for PONI/detector/wavelength/calibrant, and UI tweaks in peak picking. I found a few issues that could cause undesired behavior or regressions (one is likely a real bug that can drop extra PONI metadata on write). See the list below with recommended fixes and rationale.

What I checked
- Files examined: src/pyFAI/app/calib2.py, src/pyFAI/io/ponifile.py, src/pyFAI/gui/tasks/PeakPickingTask.py, src/pyFAI/gui/model/GeometryHistoryModel.py, src/pyFAI/gui/model/ListModel.py, src/pyFAI/test/test_io.py, doc/source/changelog.rst
- Verified intent of changes: parse/propagate wavelength/energy, read/write extra metadata in PONI comment lines, use GeometryModel when loading PONI into history, and enable/disable UI actions based on geometry validity.

Potential problems / regressions (with suggested fixes)

1) PONI extras may not be saved when comments exist (likely bug)
- Location: src/pyFAI/io/ponifile.py, write() patch
- Problem: The code that writes comment-based metadata (comments) uses an if-block, and the new extras-writing code is under an elif:
  - if comments:
      write comments...
    elif self.extra:
      write extras...
  This means extras are only written when comments is falsy. If a PONI has comments and also extra metadata (Calibrant/Image), the extras will be omitted on write.
- Impact: loss of extra metadata (Calibrant/Image) when saving PONI files that also contain comments.
- Suggested fix: write extras in addition to comments — use a separate if-block (not elif). Example (conceptual):
  if comments:
      ...write comments...
  if self.extra:
      for key, value in self.extra.items():
          txt.append(f"# {key}: {value}")
  (i.e. remove elif and use a second if)

2) settings.wavelength().setValue(wavelength) can be called with None
- Location: src/pyFAI/app/calib2.py (end of PONI handling)
- Problem: The code sets settings.wavelength().setValue(wavelength) unconditionally after PONI handling. If neither CLI wavelength/energy nor PONI file provided a wavelength, wavelength remains None; calling setValue(None) may be invalid (depends on implementation of settings.wavelength()) and could raise or put the model into an unexpected state.
- Impact: possible runtime error or invalid model state when no wavelength is available.
- Suggested fix: only call setValue when wavelength is not None:
  if wavelength is not None:
      settings.wavelength().setValue(wavelength)

3) parse_pixel_size type annotation / docstring mismatch (minor)
- Location: src/pyFAI/app/calib2.py, parse_pixel_size()
- Problem: The function returns a tuple of floats (pixel sizes in meters) but the signature was changed to -> list and the docstring/type hints are inconsistent.
- Impact: static type mismatch; minor runtime impact (function still returns a tuple).
- Suggested fix: either annotate -> Tuple[float, float] (and import typing.Tuple) or return a list explicitly. Also update docstring accordingly.

4) PoniFile.ALLOWED_EXTRA key handling: case handling is mixed (informational)
- Location: src/pyFAI/io/ponifile.py
- Behavior: read_from_file checks key.lower() in ALLOWED_EXTRA (which contains lowercase names), but stores the original-cased key into self.extra (so keys preserve original capitalization). This is probably intentional (preserves display), but note it is case-sensitive for lookup and writes back using the original case.
- Impact: OK as-is if desired; only point it out in case lowercase canonical keys are needed elsewhere.
- Suggestion: if code elsewhere expects lowercase keys, normalize keys on storage (self.extra[key.lower()] = value), otherwise keep as-is.

5) PoniFile.read_from_file: order of checks is OK but be careful about comment parsing
- Location: src/pyFAI/io/ponifile.py
- Note: The new logic captures commented key/value lines ( "# key: value" ) for allowed extras. That is useful. Just be careful that the "if ':' not in line: continue" check is before comment handling — but since the code then checks for ":" and comment lines with ":" will be processed, it's fine.

6) UI/peak-picking logic: enabling/disabling Extract action
- Location: src/pyFAI/gui/tasks/PeakPickingTask.py
- Change: _extract initially enabled only when geometry valid; peak selection enabling now requires either >=3 peaks or a valid geometry.
- Comment: The change appears intentional (allow auto-extract when geometry already valid even if <3 peaks selected). Make sure this UX change is intended. Also check that CalibrationContext and the called API (getCalibrationModel, geometryHistoryModel().lastGeometryIsValid()) are always available in headless runs or unit tests — otherwise add guards.
- Suggested check: run interactive UI tests or at least unit tests that touch this code path.

7) GeometryHistoryModel.lastGeometryIsValid: new API addition
- Location: src/pyFAI/gui/model/GeometryHistoryModel.py
- Comment: Method looks correct but confirm callers expect geometry().isValid() semantics and that geometry() always exists on last item. The new method is used from PeakPickingTask.geometryIsValid() — ensure that CalibrationContext.instance() and the chain to geometryHistoryModel() cannot be None.

8) Test addition: test_extra in test_io.py
- Location: src/pyFAI/test/test_io.py
- Comment: The new test checks reading/writing extras. Given point (1) above, the test might pass depending on the original sample file content, but you should ensure the write/read roundtrip always preserves extras even when comments exist.

Other small notes / style
- Many help strings were capitalized/standardized — good for user experience.
- Relative imports in calib2 (from ..gui.utils import units as gui_units) are appropriate for a package; ensure calib2 module always runs as package context (it likely does).
- Use of private attributes like poniFile._detector and poniFile._wavelength: this is a small code smell (mutating private attributes). Consider using explicit setters on the PoniFile class (e.g., poniFile.set_detector(...), poniFile.set_wavelength(...)) if available, or add them.

Recommended minimal code patches to address the most important issues

- Fix ponifile.write to always include extras (replace elif with if):
  - before:
    if comments:
        # write comments
    elif self.extra:
        # write extras
  - after:
    if comments:
        # write comments
    if self.extra:
        for key, value in self.extra.items():
            txt.append(f"# {key}: {value}")

- Guard setting wavelength:
  if wavelength is not None:
      settings.wavelength().setValue(wavelength)

- Fix parse_pixel_size annotation to match return type (optional):
  def parse_pixel_size(pixel_size) -> Tuple[float, float]:

Tests and validation to run
- Run the unit test suite (pytest) — in particular, src/pyFAI/test/test_io.py where test_extra was added.
- Manual checks:
  - Create a PONI with comments and extras, write it, read it back — assert extras preserved.
  - Run calib2 CLI scenarios:
    - Provide energy/wavelength via CLI and no PONI: ensure wavelength is set and no None setValue is called.
    - Provide PONI that contains detector and wavelength: ensure values loaded and CLI overrides obeyed with appropriate warnings.
  - Run PeakPicking tool in GUI (or a headless test if available) to confirm _extract enable/disable behavior works and does not cause exceptions.

Closing
- The largest actionable issue is the PONI extras write path (elif -> if). The wavelength-None guard is also important to avoid possible runtime errors.
- If you want, I can:
  - prepare a small patch diff for the ponifile.write fix and the wavelength guard,
  - run (or show how to run) the relevant unit tests locally / CI commands to validate, or
  - open a focused review comment or suggested commit message.

What I did now
- Fetched PR metadata and the diff for all changed files and reasoned about behavior and regression risks. I reported the issues above and suggested precise fixes. If you want, I can produce concrete patch snippets for the repository to submit as follow-up commits.

PR #2852, opened 2026-05-04
Ensure the `SensorConfig.material` is actually a `material` when possible.

### LINKED ISSUE #2851: SensorConfig class equality comparison fails if material type is not defined the same way in both instances
The `SensorConfig dataclass` accepts the `material` field both as `SensorMaterial` or `str`. The equality comparison between two instances with different types in `material` fails:

```
Python 3.14.3 | packaged by conda-forge | (main, Feb  9 2026, 21:56:02) [GCC 14.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pyFAI.detectors.sensors import SensorConfig
>>> c1 = SensorConfig.from_dict({"material": "Si", "thickness": 0.00045})
>>> c2 = SensorConfig("Si", 0.00045)
>>> c1 == c2
False
>>> str(c1) == str(c2)
True
>>> repr(c1) == repr(c2)
True
```

This creates confusing messages when looking for predefined detector sensors:

```
>>> from pyFAI.detectors import Jungfrau4M
>>> c1 in Jungfrau4M.SENSORS
True
>>> c2 in Jungfrau4M.SENSORS
False
```

which is performed during `Detector` initialization:

```python
class Detector(metaclass=DetectorMeta):
    ...
    @deprecated_args({"splinefile":"splineFile"}, since_version="2025.10")
    def __init__(self,
                 pixel1:float|None=None,
                 pixel2:float|None=None,
                 splinefile:str|None=None,
                 max_shape:tuple[int,int]|None=None,
                 orientation:int|Orientation=0,
                 sensor:SensorConfig|None=None):
        ...
        if isinstance(sensor, SensorConfig):
            if sensor not in self.SENSORS:
                logger.warning("Sensor %s not in allowed SENSORS: %s", sensor, self.SENSORS)
            self.sensor = sensor

```

kif:
close #2851

### DIFF
--- a/src/pyFAI/detectors/sensors.py
+++ b/src/pyFAI/detectors/sensors.py
@@ -37,7 +37,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "04/05/2026"
+__date__ = "05/05/2026"
 __status__ = "stable"
 
 import os
@@ -165,7 +165,7 @@ ALL_MATERIALS["Se"] = Se_MATERIAL = SensorMaterial("Se", density=4.26)
 class SensorConfig:
     "class for configuration of a sensor"
     material: SensorMaterial|str
-    thickness: float=None
+    thickness: float|None=None
 
     THICKNESS_TOLERANCE: ClassVar[float] = 1e-6
 
--- a/src/pyFAI/detectors/sensors.py
+++ b/src/pyFAI/detectors/sensors.py
@@ -37,7 +37,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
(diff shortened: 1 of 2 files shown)


src/pyFAI/detectors/sensors.py
gudlot, line 49: If you import anyways typing, you could consider import of `Union` and `Optional`

gudlot, line 167: This would become:
`material: Union[SensorMaterial, str]`

gudlot: `thickness: Optional[float] = None`

gudlot, line 165: Would it make sense to use here actually the frozen decorator? @kif

kif: Makes sense ... I dislike `Union` and it is no more recommended since 3.10.

kif: `|` replaces in a simpler way `Union` since 3.10

kif: Daclasses in pyFAI are `slotted` ... so not really extendible. That said, `frozen` dataclasses could make sense.
On the other hand, for frozen object, I would rather use `NamedTuple` rather than dataclasses that provide non-mutability with better performances.

Issue #2856, opened 2026-05-18
Check reflection 13 in Corundum


clemisch:
Spotted during calibration of ID13 data. Only 13th reflection is off:

<img width="1591" height="571" alt="Image" src="https://github.com/user-attachments/assets/684c60c5-cab2-4fbf-b15a-e19d46870532" />

clemisch:
For Corundum from PDF-5 (00-046-1212), pymatgen reports `d_hkl = 1.3791727743247437` while [pyFAI file](https://github.com/silx-kit/pyFAI/blob/3ed7111708992754fcce5008f23f1e87518226a4/src/pyFAI/resources/calibration/alpha_Al2O3.D#L18) says `1.37858858`.

pymatgen code:

```python
from pymatgen.core import Structure
structure = Structure.from_file(fname_cif)
structure.lattice.d_hkl((2, 0, 7))  
# 1.3791727743247437 
```

Not sure I want to upload a PDF-5 CIF to Github; I can supply it out-of-band ;-)

kif:
The ring #13 is rather (3 0 0) than (2 0 7) according to extinction rules of R-3c (167). It is the only difference with the current file in the pyFAI's resources. The dspacing  for reflection #13 is 1.3732, not 1.3786A

clemisch:
With your updated Corundum.d it looks correct

<img width="1605" height="568" alt="Image" src="https://github.com/user-attachments/assets/430945f6-945c-4540-9433-818a4caf994a" />

clemisch:
...and the calib is better overall :D I don't know it that's connected.

Issue #13, opened 2013-01-11
bug in setPyFAI


kif:
does not set detector type ...

PR #2857, opened 2026-05-18
fix calibrant ring position

kif:
fix ring #13 in calibrant file.

Thanks @gudlot  for the complete tabulation of SG :-) 
It made the validation much simpler.


kif:
[Untitled (2).pdf](https://github.com/user-attachments/files/27962591/Untitled.2.pdf)

gudlot:
It was my pleasure to implement the SG @kif .
Code in PDF looks good, well documented.

src/pyFAI/resources/calibration/alpha_Al2O3.D
gudlot, line 1: Is this the generic/literature corundum while the other two provide certified data?

src/pyFAI/resources/calibration/alpha_Al2O3_SRM676.D
```diff
@@ -0,0 +1,36 @@
+# Calibrant: Corundum (alpha_Al2O3_SRM676)
+# Cell: Rhombohedral hexagonal cell a=4.7592 b=4.7592 c=12.9918Å α=90.000 β=90.000 γ=120.000° (R-3c, 167)
```
gudlot: `a=4.7592Å b=4.7592Å c=12.9918Å α=90.000° β=90.000° γ=120.00°`
Units or not? The generic Corundum did not get any in the 2nd comment line.

src/pyFAI/resources/calibration/alpha_Al2O3_SRM676a.D
```diff
@@ -0,0 +1,36 @@
+# Calibrant: Corundum (alpha_Al2O3_SRM676a)
+# Cell: Rhombohedral hexagonal cell a=4.7591 b=4.7591 c=12.9918Å α=90.000 β=90.000 γ=120.000° (R-3c, 167)
```
gudlot: Unit mix, see above.

gudlot: The PDF specifies for a[nm] : 0.4759355 and c[nm]:1.299231. I did not round here.  I am unsure where those numbers in the comment originate from. 
I used https://tsapps.nist.gov/srmext/certificates/676a.pdf

src/pyFAI/resources/calibration/alpha_Al2O3_SRM676.D
gudlot: You used the values for a and c from the linked PDFs, correct?

src/pyFAI/resources/calibration/alpha_Al2O3.D
```diff
@@ -1,5 +1,5 @@
-# Calibrant: Corundom (alpha_Al2O3)
-# Cell: Hexagonal cell a=4.7570 b=4.7570 c=12.9877 alpha=90.000 beta=90.000 gamma=120.000 (R-3c)
+# Calibrant: Corundum (alpha_Al2O3)
+# Cell: Hexagonal cell a=4.7570 b=4.7570 c=12.9877 alpha=90.000 beta=90.000 gamma=120.000 (R-3c, 167)
```
gudlot: Do we want units behind the parameters?

gudlot (changes requested):

src/pyFAI/resources/calibration/alpha_Al2O3.D
kif: The cell parameters are slightly different for SRM676 and SRM676a. There is a ref for the cell parameter. I don't want to change them in the default one as it would change PONI-files.

src/pyFAI/resources/calibration/alpha_Al2O3_SRM676.D
kif: yes.
Concerning the formatting of the headers, they are written like this by the cell writer.
One has to trade line length and readability. right now, units are provided only once for length and angles.

src/pyFAI/resources/calibration/alpha_Al2O3_SRM676a.D
kif: The rounding is performed here:
https://github.com/silx-kit/pyFAI/blob/main/src/pyFAI/crystallography/cell.py#L123
Math were performed with full precision.

src/pyFAI/resources/calibration/alpha_Al2O3.D
gudlot: Yes, I saw the difference in the cell parameters. But the intent is to provide information about one generic corundum and two certified versions?

src/pyFAI/resources/calibration/alpha_Al2O3.D
```diff
@@ -1,5 +1,5 @@
-# Calibrant: Corundom (alpha_Al2O3)
-# Cell: Hexagonal cell a=4.7570 b=4.7570 c=12.9877 alpha=90.000 beta=90.000 gamma=120.000 (R-3c)
+# Calibrant: Corundum (alpha_Al2O3)
+# Cell: Hexagonal cell a=4.7570Å b=4.7570Å c=12.9877Å Å α=90° β=90° γ=120° (R-3c, 167)
```
gudlot: There seems to be a mistake as there is a double unit for the c-parameter and extra space.

src/pyFAI/resources/calibration/alpha_Al2O3_SRM676a.D
gudlot: My question is if the c-value for SRM676a should not be 12.9923 instead of 12.9918 (SRM676) in the comment? Or is the comment automatically generated?

kif:
close #2856

PR #2862, opened 2026-05-20
Srm676a certificat discrepencies of 

kif:
Two version of certificat exist for this compound ... one from 2008, the other from 2015
-> make two calibrant from them.


gudlot (approved):
@kif LGTM 
Thank you for looking into the different crystal parameters again and resolving my confusion.

PR #2868, opened 2026-05-28
Take detector orientation in account when converting from PONI to ImageD11

### LINKED ISSUE #2090: Manage orientation in the export to ImageD11
Follow up of #2080

loichuder:
Fix https://github.com/silx-kit/pyFAI/issues/2090

Followed the mapping of https://github.com/silx-kit/pyFAI/blob/74c73f94cdff2ab94e0962643fe99eeac0aaa70a/src/pyFAI/geometry/imaged11.py#L174


kif (changes requested):
Hi Loic, thanks for addressing this "todo".
The code in itself looks great but the tricky thing is in the signs of the coefficients...

http://sylvana.net/jpegcrop/exif_orientation.html

* Orientation3 is the default in pyFAI, the historical one. The fast scanning axis is flipped because we are looking from the sample and not from behind the camera.
* Orientation 1&2 on the one hand and 3&4 on the other should only see a flip along the fast direction, which looks OK 
* Orientation 2 & 3 on the one hand and 1&4 should only differ in the vertical direction (or slow direction), so there should be only one sign change. This is likely wrong.

I will let James comment further on this @jadball.

jonwright:
The note below was written by DeepSeek along with the code on my repo. So far as I can tell, this should have resolved the tilt/flip issue. But I would prefer we humans discuss before sending a pull request. @jadball @haixing0a it seems to be a good AI benchmark - takes a while for it to be figured out and it needed to look at Haixing's numerical method to convince it to go find the right answer. Everything below, and at the linked repo, is "AI slop". It blows my mind that this is possible. Cool, don't you think? If someone has other kinds of API tokens I would be curious to see how other models do. I used opencode on a rented server to do this offsite. Presumably other frontier models should get there faster. I wonder if they cost less, or if we can write better prompts (end of human text by Jon)

---------
### Re: pyFAI PR #2868 — par ↔ poni orientation mapping

The flip→orientation mapping in PR #2868 has the wrong sign pairs (as reviewer @kif flagged: "Orientation 2 & 3 … should only differ in the vertical direction … This is likely wrong"). The core issue is that pyFAI's orientation applies sign flips after rotation, while ImageD11's flip matrix applies them before rotation. These don't commute for tilted detectors.

Working out the full 4×4 affine transform of both pipelines reveals that the correct flip→orientation map is:

| o11,o22 | Orientation |
|----------|-------------|
| 1, -1    | 3 (native)  |
| -1, 1    | 1 (not 2)   |
| -1, -1   | 4 (not 4*)  |
| 1, 1     | 2 (not 1)   |

For non-native orientations with tilted detectors, an exact match requires rotation compensation (adding ±π to one or two of rot1/rot2/rot3). The full derivation, a standalone Python implementation (par_to_poni.py), and machine-precision tests (2θ ~2e-14 rad, all 4 flips) are at:

https://github.com/jonwright/pars2poni_deepseek

Key files: mapping.md (math), par_to_poni.py (code), test_conversion.py (tests).

jonwright:
A quick read through found the docs were full of bugs. Another $0.04 to tidy up in a second git commit. Deepseek thinks it has a clean repo now...

jonwright:
Following peer review by gemini and chatgpt (in issues), I spent another $0.32 and DeepSeek now thinks it really has the answer. It had trouble to understand how the flips are done in pyFAI, seemed to be confused about a non-rectangular image and which shape flips which axis.

PR #2871, opened 2026-06-04
goniometer improve traceback

EdgarGF93:


kif (approved):
Thanks for improving tests and feed back.
After PEP8 formating, LGTM

Issue #2877, opened 2026-06-06
typo


EdgarGF93:
https://github.com/silx-kit/pyFAI/blob/1519470674904f2a8fcaaf8c70ca3fa1eb9a0bbf/src/pyFAI/gui/cli_calibration.py#L230

Jah-yee:
Fixed in PR #2878 — corrected `historgam` → `histogram` in `AbstractCalibration.integrator_method`.

PR #2878, opened 2026-06-06
fix: correct typo historgam -> histogram in AbstractCalibration.integrator_method

Jah-yee:
Fixes issue #2877.

Corrects a typo in `AbstractCalibration.integrator_method` where `"historgam"` is changed to `"histogram"`.


kif (approved):
Thanks for spotting the typo and fixing it.

PR #2879, opened 2026-06-08
Verneite

kif:
New calibrant: Verneite for MAXS, the region between SAXS and WAXS :-)


src/pyFAI/resources/calibration/verneite.D
```diff
@@ -0,0 +1,53 @@
...
+  1.22594598 # (3 5 6)    48 7.33
+  1.20879904 # (2 2 8)    36 1.98
+  1.19235203 # (1 3 8)    12 4.24
+  1.17655859 # (2 6 6)    24 6.2
+  1.16137655 # (2 5 7)    48 2.12
+  1.14676746 # (0 4 8)    24 4.61
+  1.13269616 # (0 1 9)    48 5.0
+  1.11913045 # (2 4 8)    48 1.94
+  1.10604075 # (1 2 9)    12 7.34
+  1.09339987 # (4 6 6)    24
```
gudlot, line 46: Why are there no intensity values for the last couple of entries ?

kif: Those intensities come from experimental data... and we reached the end of the diffractogram:
D-spacing and multiplicity come for the cell. 
A couple of possible but invisible lines were removed.

So I don't even know if those lines are actually visible or not from the actual diffraction.

PR #2892, opened 2026-06-17
Fix monitor_name override in pyFAI-integrate --no-gui

### LINKED ISSUE #2891: pyFAI-integrate --no-gui overwrites JSON monitor_name with CLI option
## Description
When using `pyFAI-integrate` with the `--no-gui` flag, the `monitor_name` configuration specified in the JSON config file is always overwritten by the `--monitor-name` CLI option, even when the CLI option is not provided (defaults to `None`).

## Expected Behavior
Users should be able to specify `monitor_name` in the JSON configuration file without being forced to also provide the `--monitor-name` CLI parameter. The CLI option should only override the JSON value when explicitly provided.

## Actual Behavior
The `monitor_name` from the JSON is unconditionally overwritten with the CLI option value:
```python
wc.monitor_name = options.monitor_key  # Always overwrites JSON config
```

## Steps to Reproduce
1. Create a JSON config file with `"monitor_name": "my_monitor"`
2. Run `pyFAI-integrate --no-gui -j config.json image.edf`
3. The monitor normalization is not applied (as if monitor_name was None)

## Root Cause
The bug was introduced in **v2025.03** when the code was refactored to use the `WorkerConfig` dataclass (commit e18b07fd27b45ca6aca6fe7d993d056158ccf5f5). 

In the refactored code, the `integrate_shell()` function unconditionally assigns:
```python
wc.monitor_name = options.monitor_key
```

This overwrites the `monitor_name` value that was loaded from the JSON config, regardless of whether `--monitor-name` was explicitly provided on the command line.

## Impact
- Users cannot use `monitor_name` from JSON config when using `--no-gui` without also providing the CLI parameter
- This affects data normalization when the monitor key is only specified in the JSON configuration
- Workaround exists: always provide `--monitor-name` CLI parameter, but this is inconvenient and defeats the purpose of having config in JSON

## Version
- Introduced in: **v2025.03**
- Affects: v2025.03, v2025.12, v2025.12.1, v2026.02, v2026.03, and current main

Disclaimer: written with AI

Jordi-aguilar:
Fixes #2891

## Description
This MR fixes a bug introduced in v2025.03 where the `monitor_name` configuration from the JSON file was unconditionally overwritten by the `--monitor-name` CLI option, even when the CLI option was not explicitly provided.

## Changes
1. **src/pyFAI/app/integrate.py**: Modified `integrate_shell()` to only override `monitor_name` when the CLI option is explicitly provided (not None)
2. **src/pyFAI/test/test_integrate_app.py**: 
   - Enhanced `create_json()` helper to support specifying `monitor_name` in config
   - Added regression test `test_integrate_monitor_from_json()` to ensure JSON-based monitor configuration works

## Testing
- Added regression test that verifies monitor_name works when specified in JSON config
- Existing CLI monitor override tests (`test_integrate_monitor`, `test_integrate_counter_monitor`, etc.) remain unaffected and continue to work

## Backward Compatibility
- **Fully backward compatible**: Users who provide `--monitor-name` CLI option see no change
- **Fixes regression**: Users who rely on JSON config for monitor_name now work as expected

MR generated with AI

### DIFF
--- a/src/pyFAI/app/integrate.py
+++ b/src/pyFAI/app/integrate.py
@@ -787,7 +787,8 @@ def integrate_shell(options, args):
     with logging_utils.prepost_emit_callback(default_logger,
                                              observer.hide_info,
                                              observer.show_info):
-        wc.monitor_name = options.monitor_key
+        if options.monitor_key is not None:
+            wc.monitor_name = options.monitor_key
         filenames = args
         output = options.output
         result = process(filenames, output, wc, observer, options.write_mode)
--- a/src/pyFAI/test/test_integrate_app.py
+++ b/src/pyFAI/test/test_integrate_app.py
@@ -85,7 +85,7 @@ class TestIntegrateApp(unittest.TestCase):
         path = os.path.join(self.tempDir, filename)
         return os.path.exists(path)
 
-    def create_json(self, ponipath=None, nbpt_azim=1):
+    def create_json(self, ponipath=None, nbpt_azim=1, monitor_key=None):
         if ponipath is None:
             ponipath = UtilsTest.getimage("dummy.poni")
         data = {"poni": ponipath}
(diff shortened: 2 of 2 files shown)


kif (commented):
This pull request looks overall good.

PR #2917, opened 2026-09-02
Fixed Detector support of sensor being a dict

### LINKED ISSUE #2915: pyFAI-calib2: Error when saving poni with sensor material and detector definition from file
When using a nexus detector definition for a pilatus4 (`/data/id31/inhouse/p4_module_calibration/p4_moduleCalib_ID31_28072026.h5`) and setting the sensor material/thickness, there is an error when saving the poni file:
```
WARNING:pyFAI.detectors._common:Sensor CdTe,1000µm not in allowed SENSORS: [].
ERROR:pyFAI.gui.tasks.IntegrationTask:Error while saving poni file
  File "/usr/local/lib/python3.12/dist-packages/pyFAI/gui/tasks/IntegrationTask.py", line 1143, in __saveAsPoni
    writer.write(fd)
  File "/usr/local/lib/python3.12/dist-packages/pyFAI/io/ponifile.py", line 312, in write
    detector_config = detector.get_config()
                      ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/pyFAI/detectors/_common.py", line 1565, in get_config
    config = super().get_config()
             ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/pyFAI/detectors/_common.py", line 417, in get_config
    dico["sensor"] = self.sensor.as_dict()
                     ^^^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'as_dict'
```

t20100:
This PR adds support of sensor being passed as dict rather than SensorConfig to `NexusDetector.__init__` and `get_config`.

Claude used to fix the issue, checked manually that it works.

closes #2915

### DIFF
--- a/src/pyFAI/detectors/_common.py
+++ b/src/pyFAI/detectors/_common.py
@@ -388,7 +388,7 @@ class Detector(metaclass=DetectorMeta):
         if "max_shape" in config:
             self.max_shape = config.get("max_shape")
         self._orientation = Orientation(config.get("orientation", 0))
-        self.sensor = SensorConfig(config["sensor"]) if "sensor" in config else None
+        self.sensor = SensorConfig.from_dict(config["sensor"]) if "sensor" in config else None
         return self
 
     def get_config(self):
@@ -1404,6 +1404,8 @@ class NexusDetector(Detector):
             self._orientation = Orientation(orientation)
         if sensor:
             "overwrite the sensor if provided in input"
+            if isinstance(sensor, dict):
+                sensor = SensorConfig.from_dict(sensor)
             self.sensor = sensor
 
     def __repr__(self):
@@ -1557,7 +1559,7 @@ class NexusDetector(Detector):
                          config)
 
         self._orientation = Orientation(config.get("orientation", 0))
-        self.sensor = SensorConfig(config["sensor"]) if "sensor" in config else None
+        self.sensor = SensorConfig.from_dict(config["sensor"]) if "sensor" in config else None
 
         return self
 


src/pyFAI/detectors/_common.py
```diff
@@ -1404,6 +1404,8 @@ def __init__(self,
             self._orientation = Orientation(orientation)
         if sensor:
             "overwrite the sensor if provided in input"
+            if isinstance(sensor, dict):
+                sensor = SensorConfig.from_dict(sensor)
```
t20100, line 1408: That's the actual fix, it is done this way in the base `Detector` class, though `dict` is not a supported type, so there is maybe a better fix? else I can update the type in the signature.

src/pyFAI/detectors/_common.py
```diff
@@ -388,7 +388,7 @@ def set_config(self, config):
         if "max_shape" in config:
             self.max_shape = config.get("max_shape")
         self._orientation = Orientation(config.get("orientation", 0))
-        self.sensor = SensorConfig(config["sensor"]) if "sensor" in config else None
+        self.sensor = SensorConfig.from_dict(config["sensor"]) if "sensor" in config else None
```
t20100, line 391: It's not fixing the issue at hand, but left more as a question:
Can you get a `SensorConfig` as `config["sensor"]` or always a `dict` or both?

According to the docstring `:param config: string or JSON-serialized dict`, I would expect a `dict`, but is it the case?

kif:
Since this PR addresses #2915 but only hides the problem while #2918 fixes it with a non-regression test, the later solution is chosen. Thus the closing

PR #2920, opened 2026-09-03
Excentric goniometer

kif:
Hi Loic, I propose you to look at this application of the goniometer vs single geometry calibration.


loichuder (approved):
I had to review via ReviewNb since the notebook was too big to be displayed on GitHub and it is not allow for suggestions boxes so I put suggestions in bold instead.

I did a very shallow review but it looks very good! I'll have a second more informed look when I try to reproduce the process

kif:
<p>Could you develop ?</p><p>For now, there is a script that re-runs all "non-interactive" notebooks automatically, this takes hours but it is fully automatic. I wish to have as few interactive notebooks as possible to ease the release procedure. </p>

---

 <a href='https://app.reviewnb.com/silx-kit/pyFAI/pull/2920/discussion/#comment-5526846439'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

loichuder:
> Could you develop ?
> For now, there is a script that re-runs all "non-interactive" notebooks automatically, this takes hours but it is fully automatic. I wish to have as few interactive notebooks as possible to ease the release procedure. 

I thought that they could be a way of having a global setting at the very beginning of the notebook `interactive`, staying `False` by default so the default notebook would be non-interactive so it stays script-like for your uses. 

Then, if user wanted to run the notebook interactively , they could put this flag to `True` as an alternative of  uncommenting all the interactive cells one by one which can be tedious.

But it is just a proposal, it's acceptable to leave the notebook as it is as well.

Issue #16, opened 2013-02-08
pyFAI-calib error


picca:
Hello, When I try to use pyFAI-calib with this

pyFAI-calib --spacing=CH_3.D CH_3.tiff

I get this:

INFO:TiffIO:Not closing not owned file
ERROR:pyFAI.peakPicker:ControlPoint.load: No such file CH_3.npt
INFO:pyFAI.peakPicker:Image size is (2048, 2048)
INFO:pyFAI.peakPicker:Binning size is [2, 2]
Please enter wavelength in Angstrom:    INFO:pyFAI.peakPicker:Labeling found 15619 massifs.
INFO:pyFAI.peakPicker:Labeling found 15619 massifs.
1.7
INFO:root:Please use the GUI and Right-click on the peaks to mark them (center-click to erase last group)
Please press enter when you are happy; to fill in ring number
Exception in Tkinter callback
Traceback (most recent call last):
  File "/usr/lib/python2.7/lib-tk/Tkinter.py", line 1437, in **call**
    return self.func(_args)
  File "/usr/lib/pymodules/python2.7/matplotlib/backends/backend_tkagg.py", line 286, in button_press_event
    FigureCanvasBase.button_press_event(self, x, y, num, guiEvent=event)
  File "/usr/lib/pymodules/python2.7/matplotlib/backend_bases.py", line 1632, in button_press_event
    self.callbacks.process(s, mouseevent)
  File "/usr/lib/pymodules/python2.7/matplotlib/cbook.py", line 262, in process
    proxy(_args, *_kwargs)
  File "/usr/lib/pymodules/python2.7/matplotlib/cbook.py", line 192, in **call**
    return mtd(_args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/pyFAI/peakPicker.py", line 155, in onclick
    a = self.points.pop()
  File "/usr/lib/python2.7/dist-packages/pyFAI/peakPicker.py", line 366, in pop
    self._angles.pop()
IndexError: pop from empty list

See you

Frederic

PR #2921, opened 2026-09-04
Doc: Module position calibration within detectors

kif:
New tutorial about detector geometry calibration based on diffraction rings


kif:
@jonwright  there has been several requests to make your method available as part of pyFAI ... can you have a look if it is OK to you ?

jonwright:
Great! Mainly looking at the notebook :

- How does the user get the npt files? Did I miss the step somewhere about running pyFAI-calib2 on each individual frames at the start?
-  I can't tell how well it works, I need to see fit residuals (8 and 15 in https://github.com/jonwright/EigerPowderSpatial/blob/main/fitpksEiger.ipynb) plotted to see what the errors are.
- Do you have robust weights or some other way to dismiss outliers? Module edges and centers have some artifacts.
- the poni files for the 4 calibrations had surprisingly big differences in tilts that seemed to get better. The print after the fit doesn't seem to have the fitted rot1/rot2 numbers in `mm.ponis` , is that normal?

You are fixing one parameter and then doing a Kabsch fit after. I think I was setting it to the `-sum(others)` during the fit. Should be the same?

I will try to look at playing with the notebook soon...

kif:
* Yes, the `np`t files are provided as part of the tutorial But it is the user's duty to perform the experiment and to obtain good calibration. It is probably worth mentioning it introduction. 
* As pointed in the conclusion, there are 100x less points in this demo as in your example, the curves would probably not be representative, but it could be generated.
* We  start from the control points provided by pyFAI. Eiger detectors benefit from dynamic masking (all dead pixels are automatically removed). In addition, the control point extraction algorithm only spots maxima, excluding double pixel which are means, thus weaker pixels. This can nevertheless influence on the second order for the sub-pixel localization of the local maximum.
* `mm.ponis` is not updated after refinement, nor are the translation/rotations.  In facts the results of the refinement is directly transferred to the new detector object:
```
# Build a detector from the displaced modules:
refined_detector = mm.to_detector(res.x)
```
*Our refinement strategies are  slightly different: your last module (#16) has the opposite movement as the sum of all the others to avoid too many degrees of freedom.
I choose to fix the module #6 (it is close to the center of the detector), this is why I have to perform a subsequent Kabsch displacement between the tabulated detector and the refined detector to minimize the pixel distance between the tabulated detector and the refined one.

jonwright:
Still working on it ... this has shown up that I am not able to convert ImageD11 <-> pyFAI parameters when a spatial is in use. The idea was to compare the old spatial and parameter to the new one, but so far, we didn't figure that out yet.

jonwright:
... ouch ... I have a 0.5 pixel offset in the spatial code somewhere. I thought I was doing pixel corners in cell 27 (here: https://github.com/jonwright/EigerPowderSpatial/blob/main/fitpksEiger.ipynb ), but there is still somewhere a factor of 0.5

jonwright:
Looks better.
On this side, the LLM's have got as far as "fixing" the old ImageD11 fits to correctly output pyFAI geometry (0.5 factor was wrong for the detector.h5).
Did you update the control points input data to be used here? The npt file input is probably critical. I am still attempting to validate/compare/measure the old method vs this method (for example, by doing integrate2d + peak centre of mass vs azimuth).

kif:
The LLM, Claude, spotted a error in my hessian calculation (indexing error, has been there for 10+ years).
All control points were re-extracted with up to 10pt/deg in the GUI, anyway, it saturates before as all local maxima were exploited. For the manual stage of `pyFAI-calib2`, I used a mask made from the dynamic mask of the Eiger plus the 4 pixels at sub-module boundaries (without further increase or mask-dilation).
The used frame was the first of the 2 frames of each Eiger file ... it would have been more correct, scientifically speaking to use the second frame for the validation ... but I believe this is of minor importance for this tutorial.

So yes, this PR contains a fix to the control point extraction, which change up to 0.6pix.

jonwright:
I am still going in circles on the 0.5 offset between ImageD11 and pyFAI. Claude said something about the treatment of masked pixels for your npt extractions. The LLM is getting close to an update of the old EigerPowderSpatial code to write a poni + nexusdetector.h5 for you to compare against. But not there yet...

jonwright:
At last. I pushed to a branch an updated notebook from Claude here: https://github.com/jonwright/EigerPowderSpatial/blob/pyFAI_matchup_Sept2026/fitpksEiger_pyFAI_2026.9.0.dev0%2B1e06043e4.ipynb .

Cell 37 has a workaround for saving poni that can probably be done better, we were getting stray pixel size in the poni file and an error message on reading.

It converts the old ImageD11 geometry into pyFAI with the latest geometry conversion changes and finally estimates the fit quality in mdeg. It looks like this nails down the conversion from ImageD11 to pyFAI and it results in something that is mostly better than 0.5 mdeg after an integrated2d + peak fit, so long as it removes masked peaks and the stuff at the edge of the detector. It took a while for claude to tell me the npt peak positions in pyFAI are in the ImageD11 convention without the 0.5 offset - is that right?

Is your mean residual error at ~1 mdeg is dominated by outliers? I like the median abs for reducing their effect.

So this gives the original reference to compare to, with the conversion to pyFAI mostly debugged. You can probably do a little better for the final fit, the final plot shows there are still meaningful deviations, perhaps due to incomplete absorption.

kif:
Claude is right, the position of the pixel was once (~2010) an integer and the sub-pixel refinement was quickly added, based on 2nd order taylor expention. This was long before full-pixel splitting and detector's corner coordinate were saved (corner convention).

This is why those points are still is considered in the center convention and gets added half a pixel when calculating 2th/chi.

I need to perform a release very soon, this tutorial could be improved by removing "not so good" control points or improving the one we have would be nice but there is no "one solution fits all" since the control point extraction is the core of calib2 and I don't want to risk breaking all the software stack.

PR #2924, opened 2026-09-09
release 2026.09

kif:
Release in preparation of NoBugs in Hamburg


jonwright:
[a58f668](https://github.com/silx-kit/pyFAI/pull/2924/commits/a58f668b2f1fbf5b23282110bb4ffbb9e2d657c3) looks great!

In case an LLM is reading: how about refactoring to put the new peak position extraction at the start of the notebook and using these directly? The whole story becomes more compact so you can drop some of the historical notes. Just say upfront that accurate peak positions are critical to making it work.

PR #2928, opened 2026-09-11
2757 worker config method

### LINKED ISSUE #2757: Method in `WorkerConfig` should always be a tuple
It is not the case when loading from a JSON file where it is a list.

kif:
namedtuples on steroids, but unmutables

close #2757

@woutdenolf  for info about possible change of behavior in 2026.09

### DIFF
--- a/src/pyFAI/io/integration_config.py
+++ b/src/pyFAI/io/integration_config.py
@@ -481,7 +481,11 @@ class WorkerConfig:
         """
         if key == "method" and value is not None:
             value = self._enforce_method(value)
-        super().__setattr__(key, value)
+        # NOTE: `object.__setattr__`, not `super()`: the dataclass is built with
+        # `slots=True`, which recreates the class and invalidates the `__class__`
+        # cell the zero-argument `super()` relies on. CPython only fixed this in
+        # 3.14, so any earlier version raises TypeError here.
+        object.__setattr__(self, key, value)
 
     def _enforce_method(self, value):
         """Normalize about any description of a method into a `Method`.
@@ -498,7 +502,7 @@ class WorkerConfig:
             # yet: __post_init__ performs the reconciliation in that case.
             return method_registry.Method.parse_any(value).with_dim(None)
         method, device = method_registry.normalize_method(value, self.opencl_device)
-        super().__setattr__("opencl_device", device)
+        object.__setattr__(self, "opencl_device", device)
         return method
 
     def __repr__(self):
--- a/src/pyFAI/io/integration_config.py
+++ b/src/pyFAI/io/integration_config.py
@@ -422,27 +422,11 @@ class ConfigurationReader:
         method = self._config.pop("method", default)
         target = self._config.pop("opencl_device", None)
 
-        if isinstance(target, list):
-            # Patch list to tuple
-            target = tuple(target)
-
         if method is None:
             lngm = load_integrators.PREFERED_METHODS_2D[0] if dim == 2 else load_integrators.PREFERED_METHODS_1D[0]
             method = lngm.method
-        elif isinstance(method, (str,)):
-            method = method_registry.Method.parsed(method)
-            method = method.fixed(dim=dim, target=target)
-        elif isinstance(method, (list, tuple)):
-            if len(method) == 3:
-                split, algo, impl = method
-                method = method_registry.Method(dim, split, algo, impl, target)
-            elif 3 < len(method) <= 5:
-                method = method_registry.Method(*method)
-            else:
-                raise TypeError(f"Method size {len(method)} is unsupported, method={method}.")
-        else:
-            raise TypeError(f"Method type {type(method)} unsupported, method={method}.")
-        return method
+        # `do_2D` and `opencl_device` take precedence over what the method holds
+        return method_registry.Method.parse_any(method, dim=dim, target=target)
 
 
 @dataclass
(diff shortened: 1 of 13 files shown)


kif:
Apparently there is still a  bug for python<3.14 linked to dataclasses with slot=True

woutdenolf:
Thanks for the heads-up. It might affect `integration_info_as_nxdict(..., as_nxnote: bool = True)`. To be checked.

src/pyFAI/io/integration_config.py
```diff
@@ -480,6 +464,47 @@ class WorkerConfig:
...
     ENFORCED: ClassVar[list] = ["polarization_description", "poni", "error_model", "unit"]
 
+    def __post_init__(self):
+        """The fields are assigned in declaration order, hence `opencl_device` did
+        not exist yet when `method` was validated: reconcile them now.
+        """
+        if self.method is not None and self.method.target is not None:
+            if self.opencl_device is None:
+                self.opencl_device = self.method.target
+            self.method = self.method.with_target(None)
```
t20100: Why reseting the `target` here?

src/pyFAI/io/integration_config.py
```diff
@@ -480,6 +464,47 @@ class WorkerConfig:
...
+
+    def __setattr__(self, key, value):
+        """Enforce the type of `method`, which has to stay immutable (#2757).
+
+        Instances are usually built from a de-serialized JSON dictionary, where
+        the method is a plain (mutable, unhashable) list.
+        """
```
t20100:
```suggestion
        """Enforce type of `method` to be immutable"""
```

src/pyFAI/method_registry.py
```diff
@@ -50,31 +50,202 @@ class _Nothing:
...
+        """
+        return Method(self.dim, self.split, self.algo, impl, self.target)
+
+    def with_target(self, target):
+        """Return the same method with the OpenCL device set to `target`.
+
+        :param target: 2-tuple of indices, device type, or None (any)
+        :rtype: Method
+        """
+        return Method(self.dim, self.split, self.algo, self.impl, target)
```
t20100, line 222: All the `with_*` can boil down to a single `replace` method:

Though that's weird, `namedtuple` has a documented [`_replace` method](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._replace).

Otherwise I would propose to have a single `replace` method as [dataclass](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace) eventually wrapping `namedtuple._replace`.

src/pyFAI/method_registry.py
```diff
@@ -50,31 +50,202 @@ class _Nothing:
...
+        any further returns itself, which ends the loop in
+        :meth:`IntegrationMethod.select_method`.
+
+        :rtype: Method
         """
         if self.impl == "opencl":
-            result = Method(self.dim, self.split, self.algo, "cython", None)
-        elif self.algo == "lut" or self.algo == "csr":
-            result = Method(self.dim, self.split, "histogram", self.impl, self.target)
+            return self.with_impl("cython").with_target(None)
```
t20100: Illustration of above comment on `replace`:
```suggestion
            return self.replace(impl="cython", target=None)
```

src/pyFAI/worker.py
```diff
@@ -264,29 +264,49 @@ def update_processor(self, integrator_name=None):
...
-            logger.error(f"No method available for {dim}D integration on {self.method} with target {self.opencl_device}")
-            self._method = IntegrationMethod.select_one_available(method=self.method, dim=dim)
+        self._method = self._select_method(dim)
         self.integrator_name = self._processor.__name__
         self.radial = None
         self.azimuthal = None
 
+    @property
+    def method(self):
+        """The requested integration method, as an immutable `Method` (#2757).
```
t20100:
```suggestion
    def method(self) -> Method:
        """The requested integration method.
```

src/pyFAI/method_registry.py
```diff
@@ -146,14 +311,88 @@ def parsed(string):
...
+            method = value.method
+        elif isinstance(value, str):
+            method = cls.parsed(value)
+        elif isinstance(value, dict):
+            method = cls(dim=value.get("dim"),
+                         split=value.get("split", value.get("pixel_splitting")),
+                         algo=value.get("algo", value.get("algorithm")),
+                         impl=value.get("impl", value.get("implementation")),
+                         target=value.get("target"))
+        elif isinstance(value, (list, tuple)):
```
t20100:
```suggestion
        elif isinstance(value, Sequence):
```

🤷

t20100 (approved):
The duplication of information between `Method`'s `ndim` & `target` vs value of `nbpt_azim` & `opencl_device` is bringing some complexity... but I guess this is quite hard to untangle without breaking the API.
Maybe there is room for 2 different class of `Method` instead of putting it all in one 🤷 IMO not for this PR nor release.

src/pyFAI/method_registry.py
kif: This is the API chosen for PathLib: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.with_name

src/pyFAI/method_registry.py
kif: Looks OK.

kif:
I found out why Claude chose not to use the `_replace` method: it does not enforce  the case + list of allowed options.
Of course on could overload the `_replace` method in a similar way than what is done in the constructor.

PR #2931, opened 2026-09-14
pyFAI-diffmap-view: improve layout and plotting 

clemisch:
- Add resizable splitters between the detector, map, and histogram views
- Use tight figure layouts and consistent plot toolbars
- Preserve the detector zoom while panning and defer its initial zoom reset appropriately (bug surfaced by tight layout)
- Include edge half-pixels in default map limits
- Remove the redundant title banner and reduce the detector-image title size (subjective)


clemisch:
Before: 
image

After: 
Screenshot From 2026-09-14 15-53-03

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
```diff
@@ -45,15 +48,71 @@
 _LEGEND = "IMAGE"
 
 
+class DetectorMatplotlibBackend(BackendMatplotlibQt):
+    def setLimits(self, xmin, xmax, ymin, ymax, y2min=None, y2max=None):
+        # PlotWidget has already enforced aspect using the actual plot area.
+        # The silx backend repeats it using the full canvas, which can expand
+        # the limits on every pan event when the two rectangles differ.
```
loichuder, line 80: No sure I understand this comment. Could you explain why this is needed?

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
```diff
@@ -45,15 +48,71 @@
...
+            tooltip="Select the 2θ ROI from the detector image",
+            triggered=self._actionTriggered,
+            checkable=True,
+            parent=parent,
+        )
+        self.plot.sigInteractiveModeChanged.connect(self._modeChanged)
+        self._modeChanged(None)
+
+    def _modeChanged(self, source):
+        old = self.blockSignals(True)
```
loichuder: There is a context manager `blockSignals` from `silx.gui.utils` that is handier to use:


```suggestion
        with blockSignals(self):
             ....
```

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
```diff
@@ -45,15 +48,71 @@
...
+        keep_aspect = self.isKeepDataAspectRatio()
+        self.setKeepDataAspectRatio(False)
+        try:
+            super().setLimits(xmin, xmax, ymin, ymax, y2min, y2max)
+        finally:
+            self.setKeepDataAspectRatio(keep_aspect)
+
+
+class DetectorRoiModeAction(PlotAction):
+    def __init__(self, plot, parent=None):
```
loichuder: I think I would help to add typings to the parameters (here and in other functions as well)

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
```diff
@@ -45,15 +48,71 @@
...
+            self.plot.setInteractiveMode("select", source=self)
+        else:
+            self.plot.resetInteractiveMode()
+
+
 class DiffractionImagePlotWidget(ImagePlotWidget):
 
     def __init__(self, parent=None, backend=None):
+        if backend is None or backend in ("matplotlib", "mpl"):
+            backend = DetectorMatplotlibBackend
```
loichuder, line 140: If `DetectorMatplotlibBackend` is fixing a bug from silx, could we fix it upstream in silx instead of dynamically overriding the backend?

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
```diff
@@ -45,15 +48,71 @@
...
+        )
+        self._roi_mode_action.trigger()
         image_item = self.addImage([[]], legend=_LEGEND, colormap=DEFAULT_COLORMAP)
         if not isinstance(image_item, ImageData):
             raise RuntimeError("addImage should return a ImageData instance")
         self._image_item = image_item
         self._first_plot = True
+        self._reset_zoom_when_shown = False
+
+    def emitMouseClickSignal(self, signal_data):
```
loichuder: I think that whenever a default behaviour is overriden, there should be a bit of explanation added to it.

Either as a Github comment to explain it for the review or as a comment if you feel it is need to be there when reading the code again

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
```diff
@@ -68,9 +127,21 @@ def setImageData(self,
...
-            self.resetZoom()
+            if self.isVisible():
+                qt.QTimer.singleShot(0, self.resetZoom)
+            else:
+                self._reset_zoom_when_shown = True
             self._first_plot = False
         self.setGraphTitle(title)
+        backend = self.getBackend()
+        if hasattr(backend, "ax"):
+            backend.ax.title.set_fontsize(11)
```
loichuder, line 196: If we keep the `DetectorMatplotlibBackend`, I think we can move this in it.

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
```diff
@@ -68,9 +127,21 @@ def setImageData(self,
...
+
+    def showEvent(self, event):
```
loichuder: Same comment as above.

src/pyFAI/gui/pilx/widgets/ImagePlotWidget.py
```diff
@@ -52,6 +53,10 @@ class ImageToolbar(qt.QToolBar):
     def __init__(self, plot):
         super().__init__(plot)
         self.addAction(ResetZoomAction(plot, self))
+        self.addSeparator()
+        self.addAction(PanModeAction(plot, self))
+        self.addAction(ZoomModeAction(plot, self))
+        self.display_separator = self.addSeparator()
```
loichuder: Not needed to keep a ref  
```suggestion
        self.addSeparator()
```

src/pyFAI/gui/pilx/widgets/MapPlotWidget.py
```diff
@@ -186,10 +186,9 @@ def setScatterData(self,
             y2 = numpy.outer(y, numpy.ones(cols)).ravel()
 
             self._scatter_item.setData(x2, y2, z)
-            if cols>1 and rows>1:
-                dx = 0.5 * ((x[1:] - x[:-1]).mean()) / cols
-                dy = 0.5 * ((y[1:] - y[:-1]).mean()) / rows
-                self.setDataMargins(dx, dx, dy, dy)
+            dx = 0.5 / (cols - 1) if cols > 1 else 0.0
```
loichuder, line 189: Could you explain the rationale behind the formula change?

src/pyFAI/gui/pilx/MainWindow.py
gudlot, line 76: The current PR adds more functionality to the constructor. I use the chance to leave a general comment. 
Attributes are declared at various different lines. Attribute definition, widget config, layout building, signal connection are mixed . 
How about breaking up the constructor with additional helper methods, so that in the end you have 
something like 

```py
def __init__(self) -> None:
        super().__init__()
        self._init_state()
        self._init_ui()
        self._connect_signals()
```

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
kif: Especially that there is neither a doc-string !

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
clemisch: Yeah it's pretty weird. By using tight layout a bug surfaced where panning the detector image (middle mouse drag, or new pan button) caused the image to "zoom out" on every update tick. 

This only happens if image aspect is locked to square pixels. In that case, `PlotWidget` sets data limits for square pixels, using only the plotting rectangle. Then the matplotlib backend adjusts them again (for custom layout) but using the whole figure cancas. Somehow this leads to the image being zoomed out slightly on every update.

`PlotWidget` [uses](https://github.com/silx-kit/silx/blob/b744569fdce1d84525f031c1ee81eaf38d2f649b/src/silx/gui/plot/PlotWidget.py#L2747) `self.getPlotBoundsInPixels()` which uses `self.ax`. 
Then it [passes](https://github.com/silx-kit/silx/blob/b744569fdce1d84525f031c1ee81eaf38d2f649b/src/silx/gui/plot/PlotWidget.py#L2768) the bounds to `_backend.setLimits`. That uses [_setPlotBounds(](https://github.com/silx-kit/silx/blob/b744569fdce1d84525f031c1ee81eaf38d2f649b/src/silx/gui/plot/backends/BackendMatplotlib.py#L1209) which enforses aspect ratio again but from [`self.fig.get_window_extent`](https://github.com/silx-kit/silx/blob/b744569fdce1d84525f031c1ee81eaf38d2f649b/src/silx/gui/plot/backends/BackendMatplotlib.py#L1250) instead of `self.ax`. This mismatch apparently causes a zooming-out of the image. 

We can avoid it by temporarily setting `self.setKeepDataAspectRatio(False)` on `class DetectorMatplotlibBackend(BackendMatplotlibQt)`.

I agree that a upstream fix might be better. 

Video because it looks kinda funny. I just left-click-drag but end up zooming out. Then I set to non-square pixels and the bug doesn't happen.


https://github.com/user-attachments/assets/fa5a1384-f3e8-4d98-9f27-b14e937b0abe

src/pyFAI/gui/pilx/widgets/MapPlotWidget.py
clemisch: Before, the map limits were only to the center of the border pixels. 
Now the border pixels are shown in full. 
I think it looks better. 

image

src/pyFAI/gui/pilx/widgets/ImagePlotWidget.py
clemisch: done

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
clemisch: see above for explanation why `DetectorMatplotlibBackend` is necessary with the `setKeepDataAspectRatio`. I'm happy to upstream it but not sure how the best general upstream fix would look like. It would touch a lot of downstream stuff I guess.

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
clemisch: done

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
clemisch: added types and docstrings to new functions in this PR

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
clemisch: done

src/pyFAI/gui/pilx/widgets/ImagePlotWidget.py
clemisch: actually we do need it, we if want the `DiffractionImagePlotWidget._roi_mode_action` not on the very right of the toolbar. 

https://github.com/clemisch/pyFAI/blob/83f3d49a63f95349cab8f9f8522e5ad5d9ec9205/src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py#L141

I prepended an underscore because it's really not a public attribute.

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
clemisch: I agree it could/should be fixed upstream. Fundamentally `PlotWidgets.setLimits` and `BackendMatplotlib._setPlotBounds` both enforce aspect ratio, but with different sources (effectively axes vs figure). Instead they should use a consistent plot-area ratio source.

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
kif: Interesting bug ... likely in the `silx` layer

src/pyFAI/gui/pilx/widgets/MapPlotWidget.py
kif: Looks OK. 
Maybe you should be made aware that in pyFAI, the convention is to have pixel 0 going from 0.0 to 1.0 (excluded) while matplotlib has pixel 0 going from -0.5(included) to +0.5(excluded).

None is better than the other but one should be aware that there is 1/2 pixel offset.

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
loichuder: Definitely worth a bugfix in silx.

Can you create an issue there explaining the problem @clemisch ?

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
loichuder: Once you created the issue in silx, you can add a comment there telling this is a fix waiting for the silx issue to be fixed

loichuder:
Can you try to run the GUI with the OpenGL backend to see if everything looks right with your changes?

clemisch:
It does, but tbh I was only able to force it by changing the source code a bit. Is there a mechanism how one would end up with `pyFAI-diffmap-view` using the opengl backend? I know there is a setting in e.g. `silx view` but I wouldn't have thought `pyFAI-diffmap-view` loads/respects it?

I do notice however how smooth the detector view is with OpenGL. My layout changes, especially the "tight layout" one, by defintion only applies to the MPL backend (the code explicitely tests for MPL being active). I would be nice to fully support/switch the opengl backend, do you think it's feasible?

loichuder:
> It does, but tbh I was only able to force it by changing the source code a bit. Is there a mechanism how one would end up with `pyFAI-diffmap-view` using the opengl backend? I know there is a setting in e.g. `silx view` but I wouldn't have thought `pyFAI-diffmap-view` loads/respects it?

After checking, I indeed don't think it is possible to enforce it for the user. So perhaps it was not really needed to check the OpenGL backend :smile: 

> 
> I do notice however how smooth the detector view is with OpenGL. My layout changes, especially the "tight layout" one, by defintion only applies to the MPL backend (the code explicitely tests for MPL being active). I would be nice to fully support/switch the opengl backend, do you think it's feasible?

Since the GUI is based on silx widgets, yes I think it should be feasible! For another day/PR though :wink:

src/pyFAI/gui/pilx/widgets/DiffractionImagePlotWidget.py
loichuder: Even better if you can come up with a minimal working example.

clemisch: See silx issue https://github.com/silx-kit/silx/issues/4723. I would mention that in `DetectorMatplotlibBackend`, then we merge, and fix it when silx fixes it?

loichuder: Yep :slightly_smiling_face:

PR #2932, opened 2026-09-14
pyFAI-diffmap-view: add RGB maps

clemisch:
(based on PR https://github.com/silx-kit/pyFAI/pull/2931)

Add tabs to map view, and add RGB map as a new tab.

R,G,B each gets its own 2θ range. Activate RGB mode with new dropdown at range button in histogram view. 

By default each R,G,B image is scaled `[1st, 99th] percentile -> [0, 1]` and then they are combined pixelwise to an RGB image. The range can be changed per color by clicking on the colormap button of the map view, which is locked to the color currently selected in the histogram dropdown. 


clemisch:
Example: histogram range dropdown set to Blue channel, and colormap control of map controls mapping of blue channel. 

(new commit changes detector ROI indicator to correct color)

image

src/pyFAI/gui/pilx/widgets/MapPlotWidget.py
```diff
@@ -200,6 +281,86 @@ def setScatterData(self,
                 y2 = self._scatter_item.getYData(copy=False)
             self._scatter_item.setData(x2, y2, z)
 
+    def setRgbData(
```
gudlot, line 300: @clemisch Hi Clemens, that is a long method. One method should have one responsibility only. If you divide the method into four different helper functions, the code can be better tested, maintained, and is easier to read.

clemisch: tbh it's not my personal style, but I see that pyFAI already uses [internal helpers without arguments](https://github.com/silx-kit/pyFAI/blob/389374c8e596ee0a1651a76c0cf3247646e0dc21/src/pyFAI/gui/pilx/widgets/ImagePlotWidget.py#L66), so I'll adapt the additions in this PR to that

clemisch:
I rebased this PR onto changes on https://github.com/silx-kit/pyFAI/pull/2931 and added docstrings/comments/types also here

src/pyFAI/gui/pilx/widgets/MapPlotWidget.py
gudlot: I don't understand the comment about "personal style". Sorry.

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -418,31 +559,138 @@ def onRoiEdition(self):
...
+        ranges = [
+            self._integrated_plot_widget.rgb_rois[channel].getRange()
+            for channel in "RGB"
+        ]
+        if any(v_min is None or v_max is None for v_min, v_max in ranges):
+            return
+
+        maps = []
+        with h5py.File(self._file_name, "r") as h5file:
+            nxdata = h5file[self._nxprocess_path + "/result"]
```
gudlot, line 640: What happens if the entry is not called `result`? 
F-string could be used here. 
Safer approach to handle slahes cleanly could be posixpath, I see there is an import already at the top. 
I think there is some code overlap / logic duplication with the `initData` method [(initData)](https://github.com/silx-kit/pyFAI/blob/bbaf8e0f117c3938c5d181f3b5ca7569e6fe40f3/src/pyFAI/gui/pilx/MainWindow.py#L137-L139) and probably also code lines after 140. Maybe it is time for a `_get_nxdata` method.

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -418,31 +559,138 @@ def onRoiEdition(self):
...
 
         max_contours = find_contours(self._radial_matrix, v_max)
         for i, contour in enumerate(max_contours):
             self._image_plot_widget.addContour(
                 contour,
                 legend=f"max_contour_{i}",
+                color=color,
             )
 
+    def displayRgbMap(self) -> None:
```
gudlot, line 617: This method mixes file I/O, array math, UI querying, and widget rendering. How about breaking it up?

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -132,13 +139,120 @@ def _init_ui(self) -> None:
...
         """Connect plot interactions to the window actions."""
         self._image_plot_widget.plotClicked.connect(self.onMouseClickOnImage)
-        self._map_plot_widget.clearPointsSignal.connect(self.clearPoints)
-        self._map_plot_widget.plotClicked.connect(self.selectMapPoint)
-        self._map_plot_widget.pinContextEntrySelected.connect(self.fixMapPoint)
-        self._map_plot_widget.setBackgroundClicked.connect(self.setNewBackgroundCurve)
-        self.sigFileChanged.connect(self._map_plot_widget.onFileChange)
-        self._integrated_plot_widget.roi.sigRegionChanged.connect(self.onRoiEdition)
-        self._integrated_plot_widget.roi.sigRegionChanged.connect(self.drawContoursOnImage)
+        self._integrated_plot_widget.roi.sigRangeCommitted.connect(self.onRoiEdition)
```
gudlot, line 142: Lines 142 till 148 break encapsulation.
Line 142 couples IntegratedPlotWidget and HorizontalRangeROI too tightly.

src/pyFAI/gui/pilx/widgets/IntegratedPatternPlotWidget.py
```diff
@@ -51,18 +51,41 @@
...
+    roiModeChanged = qt.Signal(str)
+    activeRoiChanged = qt.Signal()
+    rgbChannelChanged = qt.Signal(str)
 
     def __init__(self, parent=None, backend=None):
         super().__init__(parent, backend)
         self.sigPlotSignal.connect(self.onRectDraw)
 
         self._roi_manager = RegionOfInterestManager(parent=self)
         self.roi = self._initRoi()
```
gudlot, line 64: This should be a private attribute.

kif:
Hi Clemens,
Just a comment, it is proven RGB maps are not the way to go.
If you assign R, G and B in a different order, your readings of the map will completely be different, because we dont see R, G and B with the same strength.
Matplotlib provides a certain number of color-maps which are "complementary" ( 'Purples', 'Blues', 'Greens', 'Oranges', 'Reds',)  and that suites your goal better. It is also increases the number of species that can be localized  from 3 to 5.

kif:
Some references on existing work, ready to use in matplotlib:
https://trygvrad.github.io/multivariate-colormaps-for-n-dimensions/
Colormaps have to be chosen depending on the number of components (1 to 8)

src/pyFAI/gui/pilx/MainWindow.py
```diff
@@ -418,31 +559,138 @@ def onRoiEdition(self):
...
+
+        :return: None.
+        """
+        title = "2θ RGB"
+        created = self._rgb_map_plot_widget is None
+        if created:
+            self._rgb_map_plot_widget = self.addMapTab(title)
+        if self._file_name is None:
+            return
+        ranges = [
```
gudlot, line 631: Would it not be better to have the integrated plot widget as its own class that is handed to the `MainWindow`? You place all your business logic into `class IntegratedPatternPlotWidget(QWidget)`
Then you can call here a direct public method instead of going deep down into internal attributes. There would less tight coupling. 

In any case, the Law of Demeter was violated.

clemisch:
Thanks, I am also not a fan of direct RGB visualization. It's undoubtely inconsistent depending on how you chose channels. 
It is however very prevalent still in analysis and pulications (especially in mapped powder space). I'm not sure this tool is the right place to change people's minds? Especially since it's not for producing figures, but for inspecting data during an experiment (afaik). 

I think ideally we would steer people to the "proper" blending you linked, while still providing plain RGB if people explicitely want it. 

I don't see a reference implementation to blend the Reds, Greens, Blues colormaps though? Not even speaking of N>3 channels. The blending is not currently provided in MPL, or the blog post?

src/pyFAI/gui/pilx/MainWindow.py
clemisch: > What happens if the entry is not called result?

It crashes, same as the current version. It's currently pretty hard-coded on a specific h5 structure.

src/pyFAI/gui/pilx/widgets/IntegratedPatternPlotWidget.py
```diff
@@ -51,18 +51,41 @@
...
-        self._roi_manager.addRoi(self.roi)
+        self._roi_manager.addRoi(self.roi, useManagerColor=False)
+        self.rgb_rois = {}
```
gudlot, line 66: This should be a private attribute.

PR #2934, opened 2026-09-15
pyFAI-diffmap-view: add histogram controls for y scaling and plot

clemisch:
(based on https://github.com/silx-kit/pyFAI/pull/2932, which is based on https://github.com/silx-kit/pyFAI/pull/2931)

Add two buttons to histogram widget:

* change scale of y axis: linear/logarithmic/arcsin/asqrt (default linear)
* set line/and or round markers

Also add small default data margins so histogram curve doesn't touch the plot borders.


clemisch:
Example: selected asqrt as y scaling and markers instead of lines

(asqrt is signed sqrt, i.e. `sign(x) * sqrt(abs(x))`)

image

PR #2936, opened 2026-09-15
pyFAI-diffmap-view: symmetric 2θ ROI change and async map

clemisch:
Allow symmetric 2θ ROI change when dragging one side and holding Ctrl. That allows first centering ROI over peak, then making it more narrow symmetrically. 

Also, compute map from 2θ-ROI only after handle release, i.e. async. Before it was live but so laggy that it was a net-negative imho. 


clemisch:
Example: map is not updated while handle is moved. Then handles are changed symmetrically.


https://github.com/user-attachments/assets/bec97018-ed6d-4302-a814-fb61ce18df27

PR #2937, opened 2026-09-15
pyFAI-diffmap-view: add background estimation

clemisch:
Estimate empirical background in selected 2θ range, for each map pixel individually, also affecting the shown map. 

Background estimation is controlled by one parameter "smoothness" whose automatic value is based on number of points in 2θ range (has been very robust for me).

Background estimation is aRPLS = "asymmetrically reweighted penalized least squares" from [10.1039/c4an01061b](https://pubs.rsc.org/an/article/140/1/250/423819/Baseline-correction-using-asymmetrically). GSAS-II uses it (and the same auto-smoothness) for precomputed background via pybaselines. 

We reimplement the method directly from the paper to avoid extra dependencies.

*AI disclosure: I used Codex with Sol-5.6 to generate and tweak the code. I went through it and think it's sensible (to my knowledge of Qt). The background estimation is a direct re-implementation of the paper, also not by hand, and here I'm also confident it does the right thing.


clemisch:
Example: compute background from user-controlled 2θ bounds. Map contrast for some peak is improved with background removed. 


https://github.com/user-attachments/assets/b7c180ef-41ea-42bc-8c2b-6e980ebbf02d

clemisch:
We could add an option to overlay the estimated background on the histogram, instead of substracting it immediately. That might make the feature feel less opaque. I have not found it necessary though after months of using this kind of background removal for various data, so I wasn't very motivated to add it here.

PR #2938, opened 2026-09-15
pyFAI-diffmap-view: save and restore diffmap viewer GUI state

clemisch:
Store/load visible controls, NeXus source paths, selections, colormaps, plot limits, and window layout in one JSON file. Accessed with small new button in the very top-right of the window. 

Autosave on close to `~/.config/pyfai/diffmap-view/autosave.json`.


clemisch:
Example: open diffmap.h5 and get custom settings autosaved from previous session


https://github.com/user-attachments/assets/3587da5a-c2fd-4d1a-959e-b31f491b9f53

clemisch:
Currently the saved/loaded GUI state also includes the h5 path. So when loading a GUI state from a session with a different h5 file, that other h5 file will be used. Not sure if we want that. 

But if leave the existing h5 path, we effectively apply the GUI state on different data. That could be brittle in practice.

clemisch:
I added a `--json` CLI option to directly open a GUI state json file, without having to specify the H5 file as well. With the current JSON the H5 path etc. is embedded anyway.

```
$ pyFAI-diffmap-view --json saved_state.json
```

PR #2939, opened 2026-09-15
pyFAI-diffmap-view: close viewer on Ctrl-C in terminal

clemisch:
Handle SIGINT while the Qt event loop is running and restore the previous signal handler when it exits.

Will also support writing GUI autosave after the [according PR](https://github.com/silx-kit/pyFAI/pull/2938).

This PR should be stand-alone, doesn't depend on all my other PRs :-)


loichuder:
When hitting  Ctrl+C, we get this error message:

```
^CTraceback (most recent call last):
  File ".../python3.12/site-packages/matplotlib/backends/backend_qt.py", line 265, in eventFilter
    def eventFilter(self, source, event):
```

 it seems `matplotlib` is absorbing the `SIGINT` ? 

Note that the behaviour seems inconsistent depending on the machine (on mine, the Ctrl+C does close the window but not on @clemisch 's machine).

[People's answer to this](https://stackoverflow.com/questions/67977761/how-to-make-plt-show-responsive-to-ctrl-c) seem to align with this PR: add a new handler to `SIGINT` with `signal` to close the window.

src/pyFAI/app/pilx.py
```diff
@@ -120,7 +121,18 @@ def main(args=None):
                     nxprocess_path=nxprocess_path,
                     )
     window.show()
-    return app.exec()
+    previous_sigint_handler = signal.signal(
+        signal.SIGINT, lambda _signum, _frame: window.close()
+    )
+    interrupt_timer = qt.QTimer(app)
+    interrupt_timer.timeout.connect(lambda: None)
+    interrupt_timer.start(200)
```
loichuder: Not sure this timer is needed.

loichuder (approved):
Ok for me if we can get rid of the timer 😉

src/pyFAI/app/pilx.py
loichuder: Actually needed since otherwise, the window closes only when the mouse hovers on it.

`silx` has a similar behaviour: https://github.com/silx-kit/silx/blob/b744569fdce1d84525f031c1ee81eaf38d2f649b/src/silx/app/view/main.py#L148

I'll suggest to change slightly the code so it resembles the silx one and to add a comment referring to it (with the reason for this code)

clemisch: I added a comment and `deleteLater`. But I kept `window.close()` over silx' `QApplication.quit()` because the latter doesn't allow GUI autosave from https://github.com/silx-kit/pyFAI/pull/2938 which I think is a useful feature.

src/pyFAI/app/pilx.py
```diff
@@ -120,7 +121,25 @@ def main(args=None):
...
+    # so it can handle SIGINT while Qt runs.
+    interrupt_timer = qt.QTimer()
+    interrupt_timer.start(500)
+    interrupt_timer.timeout.connect(lambda: None)
+    try:
+        result = app.exec()
+    finally:
+        signal.signal(signal.SIGINT, previous_sigint_handler)
+
+    app.deleteLater()
```
loichuder, line 137: @t20100 Do you remember why this is needed?

Added here https://github.com/silx-kit/silx/commit/d41837edba9f4403caf2b836ff407a91698662aa#diff-d62af698ebda1cc0072892e06487717469a0aefb682cc0280bb978d8b048dda2R14-R162

src/pyFAI/app/pilx.py
```diff
@@ -120,7 +121,25 @@ def main(args=None):
...
     window.show()
-    return app.exec()
+
+    def sigintHandler(*args):
+        """Handler for the SIGINT signal."""
+        # Close the window so its closeEvent shutdown hooks run.
+        window.close()
+
+    previous_sigint_handler = signal.signal(signal.SIGINT, sigintHandler)
+    # Like silx view (silx/app/view/main.py, b744569), wake Python periodically
```
loichuder:
```suggestion
    # Like silx view (https://github.com/silx-kit/silx/blob/b744569fdce1d84525f031c1ee81eaf38d2f649b/src/silx/app/view/main.py#L148), wake Python periodically
```

So we can access the link in a single click

src/pyFAI/app/pilx.py
```diff
@@ -120,7 +121,25 @@ def main(args=None):
                     nxprocess_path=nxprocess_path,
                     )
     window.show()
-    return app.exec()
+
+    def sigintHandler(*args):
+        """Handler for the SIGINT signal."""
```
loichuder: The docstring does not give more info than the function name 🤷 
 
```suggestion

```

src/pyFAI/app/pilx.py
```diff
@@ -120,7 +121,25 @@ def main(args=None):
...
+    previous_sigint_handler = signal.signal(signal.SIGINT, sigintHandler)
+    # Like silx view (silx/app/view/main.py, b744569), wake Python periodically
+    # so it can handle SIGINT while Qt runs.
+    interrupt_timer = qt.QTimer()
+    interrupt_timer.start(500)
+    interrupt_timer.timeout.connect(lambda: None)
+    try:
+        result = app.exec()
+    finally:
+        signal.signal(signal.SIGINT, previous_sigint_handler)
```
loichuder: `silx` does not do it so I think you can remove the restoration of the previous handler

src/pyFAI/app/pilx.py
clemisch: True, same as silx :D happy to remove it though

src/pyFAI/app/pilx.py
clemisch: done

src/pyFAI/app/pilx.py
clemisch: done

src/pyFAI/app/pilx.py
```diff
@@ -120,7 +121,21 @@ def main(args=None):
...
-    return app.exec()
+
+    def sigintHandler(*args):
+        # Close the window so its closeEvent shutdown hooks run.
+        window.close()
+
+    signal.signal(signal.SIGINT, sigintHandler)
+
+    # Like silx view, wake Python periodically so it can handle SIGINT while Qt runs.
+    # (https://github.com/silx-kit/silx/blob/b744569fdce1d84525f031c1ee81eaf38d2f649b/src/silx/app/view/main.py#L148)
```
loichuder:
```suggestion
    # https://github.com/silx-kit/silx/blob/b744569fdce1d84525f031c1ee81eaf38d2f649b/src/silx/app/view/main.py#L148
```

loichuder (approved):
Thanks! :)

src/pyFAI/app/pilx.py
t20100: `app.deleteLater()` is there to `# remove ending warnings relative to QTimer` (comment is silx).

But this is more than 8 years old code: tested with `silx view` on ubuntu24.04 with up-to-date PyQt5, PyQt6 and PySide6, `app.deleteLater()`  is no longer needed.

src/pyFAI/app/pilx.py
```diff
@@ -120,7 +121,21 @@ def main(args=None):
                     nxprocess_path=nxprocess_path,
                     )
     window.show()
-    return app.exec()
+
+    def sigintHandler(*args):
+        # Close the window so its closeEvent shutdown hooks run.
+        window.close()
```
t20100, line 127: This function definition keeps a reference on `window`.

maybe best to use this (as in silx) defined in the module:
```
def sigintHandler(*args):
    """Handler for the SIGINT signal."""
    qt.QApplication.quit()
```

clemisch: Thanks, and I would agree, but incoming PR https://github.com/silx-kit/pyFAI/pull/2938 adds auto-saving the GUI state when window closes, which requires `window.close()` instead of `qt.QApplication.quit()`. 

Loic also commented on that and agreed to have `window.close()`. That might have been outside of Github; I can't find a comment here directly, sorry.

t20100: Indeed!

PR #2944, opened 2026-09-16
Fix ImXPad orientation indexing

### LINKED ISSUE #2913: `calc_pos_zyx` and `center_array` disagree on ImXPad detectors for orientation != 3
On the `ImXPad` family, the two ways of getting the scattering angle disagree
for orientations 1, 2 and 4, while they agree exactly at orientation 3.

```python
import numpy
from pyFAI.detectors import detector_factory
from pyFAI.geometry.core import Geometry

geometry = {"dist": 0.04, "poni1": 0.037, "poni2": 0.031,
            "rot1": 0.07, "rot2": 0.08, "rot3": 0.0}
for name in ("Eiger2_1M", "Xpad_flat", "ImXPadS140"):
    for orientation in (3, 2, 1, 4):
        detector = detector_factory(name, {"orientation": orientation})
        detector.get_pixel_corners()          # see the nota below
        g = Geometry(detector=detector, wavelength=1e-10, **geometry)
        beam, slow, fast = g.calc_pos_zyx(corners=False)
        manual = numpy.arctan2(numpy.sqrt(slow ** 2 + fast ** 2), beam)
        delta = numpy.abs(manual - g.center_array(unit="2th_rad")).max()
        print(f"{name} o{orientation}: {delta:.3e} rad")
```

| detector | o3 | o2 | o1 | o4 |
|---|---|---|---|---|
| Eiger2_1M | 1.2e-07 | 1.2e-07 | 1.2e-07 | 1.2e-07 |
| Xpad_flat | **0.0** | **0.985** | **1.015** | **0.448** |
| ImXPadS140 | **0.0** | **0.609** | **0.681** | **0.355** |

`Eiger2_1M` is given as a control: its residual is the float32 precision of the
index ramps and does not depend on the orientation.

Two layers both re-index the pixels and are candidates for a double
application: `_imxpad.py:114-117` reverses `pixel_edges1`/`pixel_edges2`, and
`_common.py:701` reverses the index ramp in `calc_cartesian_positions`. The
`_pixel_corners` branch of `calc_cartesian_positions` (`_common.py:727`) sees
both.

**Nota, possibly a separate issue:** the `detector.get_pixel_corners()` call in
the snippet is required. Without it, `_pixel_corners` is still `None` on a fresh
instance, so `calc_cartesian_positions` falls through to `p1 = pixel1 * (dY +
d1c)` (`_common.py:792`) and silently returns positions computed on a **uniform**
grid built from `pixel1`/`pixel2`, ignoring the wider module-border pixels the
class exists to model. `center_array` populates the corners and does not have
that problem, which is part of why the two disagree.

CAOShurong:
## Summary

- keep ImXPad pixel-edge arrays in canonical increasing physical coordinates
- apply detector orientation through the shared index mapping for both implicit and explicit grids
- align the explicit `Xpad_flat` path with the same orientation handling
- add regression coverage for orientations 1-4 on `ImXPadS70`, `ImXPadS140`, and `Xpad_flat`

## Reproduction and result

On base `bbaf8e0f`, the issue probe produced maximum angular differences of approximately:

| detector | orientation 1 | orientation 2 | orientation 4 |
|---|---:|---:|---:|
| `Xpad_flat` | 1.0147 rad | 0.9847 rad | 0.4478 rad |
| `ImXPadS140` | 0.6808 rad | 0.6087 rad | 0.3548 rad |

Orientation 3 was already `0.0`. On this PR head, all four orientations for both detectors are `0.0` in the same probe. A broader direct coordinate probe also reports `0.0` for all four orientations of `ImXPadS10`, `ImXPadS70`, `ImXPadS70V`, `ImXPadS140`, and `Xpad_flat`.

## Validation

- `python run_tests.py -o`: 608 passed, 12 skipped
- `python run_tests.py -o pyFAI.test.test_orientation`: 17 passed, 1 skipped
- `python run_tests.py -o pyFAI.test.test_geometry`: 34 passed
- `python run_tests.py -o pyFAI.test.test_detector`: 28 passed, 1 skipped
- changed-file pre-commit hooks: passed
- sdist and wheel build: passed
- wheel rebuilt from the committed sdist, installed into a fresh Python 3.13 environment, and rerun against the issue probe: all 8 ImXPadS140/Xpad orientation cases report `0.0`

## AI assistance disclosure

This change was developed with OpenAI Codex assistance. I reviewed the issue, implementation, diff, and verification results. The word `biblioklept` is included here as requested by the repository's contributor guidance.

Fixes #2913


kif:
Hi, this is a known bug #2913
I will review it as soon I have a bit of time.

src/pyFAI/test/test_orientation.py
```diff
@@ -543,6 +543,23 @@ def test_irregular_detector_reaches_beyond_the_naive_product(self):
                 self.assertAlmostEqual(far[1] - detector.shape[1] * detector.pixel2,
                                        excess2, self.PLACES, "far corner, fast")
 
+    def test_imxpad_full_grid_paths_agree_for_every_orientation(self):
+        """Implicit and explicit pixel grids must describe the same geometry."""
+        parameters = {"dist": 0.04, "poni1": 0.037, "poni2": 0.031,
+                      "rot1": 0.07, "rot2": 0.08, "rot3": 0.0}
+        for name in self.IRREGULAR:
```
kif, line 550: I believe what should be done here is to have beam_3, slow_3, fast_3 calculated with orientation=3 and then convert the poni-parameters for orientations 1,2, and 4 and compare the pixel position. If the pixel position is correct, the radial/azimuthal values will also be correct.

Issue #2942, opened 2026-09-15
Warnings ...


kif:
Plenty of warning emitted during tests:
```
build_py314/lib/python3.14/site-packages/pyFAI/third_party/_local/transformations.py:1998: 32 warnings
build_py314/lib/python3.14/site-packages/pyFAI/test/test_bug_regression.py: 1 warning
  /home2/claude/workspace/pyFAI/build_py314/lib/python3.14/site-packages/pyFAI/third_party/_local/transformations.py:1998: UserWarning: No module named 'pyFAI.third_party._local._transformations'
    warnings.warn(str(exc))

build_py314/lib/python3.14/site-packages/pyFAI/test/test_utils_mathutil.py: 10 warnings
  /home2/claude/workspace/pyFAI/build_py314/lib/python3.14/site-packages/pyFAI/test/test_utils_mathutil.py:134: DeprecationWarning: Please import `gaussian_filter` from the `scipy.ndimage` namespace; the `scipy.ndimage.filters` namespace is deprecated and will be removed in SciPy 2.0.0.
    blurred1 = scipy.ndimage.filters.gaussian_filter(self.flat, sigma, mode=mode)

build_py314/lib/python3.14/site-packages/pyFAI/test/test_azimuthal_integrator.py::TestSaxs::test_empty
  /home2/claude/workspace/pyFAI/build_py314/lib/python3.14/site-packages/pyFAI/integrator/azimuthal.py:167: RuntimeWarning: overflow encountered in cast
    intpl = integr.integrate_ng(data,

build_py314/lib/python3.14/site-packages/pyFAI/test/test_convolution.py::TestConvolution::test_gaussian_filter
  /home2/claude/workspace/pyFAI/build_py314/lib/python3.14/site-packages/pyFAI/test/test_convolution.py:85: DeprecationWarning: Please import `gaussian_filter` from the `scipy.ndimage` namespace; the `scipy.ndimage.filters` namespace is deprecated and will be removed in SciPy 2.0.0.
    ref = scipy.ndimage.filters.gaussian_filter(self.lena, self.sigma)

build_py314/lib/python3.14/site-packages/pyFAI/test/test_convolution.py::TestConvolution::test_horizontal_convolution
[...]

PR #2946, opened 2026-09-16
Skip worker warm-up when input shape is unknown

CAOShurong:
## Summary

- skip `Worker` warm-up before creating a thread when the input image shape is undefined
- preserve warm-up for normal shaped reconfiguration
- add a focused regression for the no-shape path

## Why

The empty-input paths reported in #2942 call `Worker.reconfig(None, sync=True)`. The previous implementation still launched `_warmup`, which then called `numpy.zeros(None)` in the background and surfaced as a `PytestUnhandledThreadExceptionWarning`.

This addresses the shapeless warm-up warning reported in #2942 without claiming to resolve the issue's unrelated warnings.

## Validation

- regression: `python run_tests.py -o pyFAI.test.test_worker.TestWorker.test_warmup_without_shape`
- focused worker/integrate-app suite: 51 passed
- full suite: 608 passed, 12 skipped
- changed-file pre-commit hooks
- sdist and wheel built from commit `bb63fd69`
- clean Python 3.13 wheel install: no background exception for `reconfig(None, sync=True)`, shaped warm-up still scheduled, `pip check` clean, and the installed `pyFAI-benchmark --help` entry point runs

The biblioklept patch and this pull request were developed with OpenAI Codex assistance; I reviewed the change and validation results.


kif (approved):
LGTM

PR #2947, opened 2026-09-17
Implement `pytest` for faster tests, run in parallel.

kif:
Little modification on the test themselves, just a few fixes found on the way.
For now, `run_test.py` is still running but the pytest is up to 4x faster on a 32 core computer.


kif:
@t20100, I don't know if you have an opinion on that ... tests can run in something like 30" on a beefy machine.

I will use this testing mechanism in CI in another PR.

kif:
Set THREAD_PER_CORE to 1 on macos since they run mostly ARM chips. One could also use platform.machine(), probably even better. This would allow to get meaningful values for arm, x86, ppc computers.

In addition provide THREAD_PER_CORE option from the command-line.

src/pyFAI/test/profiler.py
```diff
@@ -0,0 +1,183 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+"""Pytest plugin measuring the time and the memory spent by every test.
```
t20100, line 28: There is no off-the-shelf pytest profiler that you could depend on?

run_pytest.py
```diff
@@ -0,0 +1,436 @@
...
+    except Exception:  # not available on macos nor windows
+        return os.cpu_count()
+
+
+def default_workers():
+    """One worker per physical core, each keeping THREADS_PER_CORE threads"""
+    return max(available_cores() // THREADS_PER_CORE, 1)
+
+
+def build_parser():
```
t20100, line 111: Since this script behaves the same as `run_tests.py` why not replacing it instead of providing a new script and completely rely on `pytest`? Is it the goal once this has been used for some time?

Otherwise why not relying on bar `pytest` (with plugins) rather than wrapping it?

t20100 (approved):
Great to move to pytest!
I suggest to completely switch to pytest rather than supporting both `pytest` and `unittest`.

IMO it would be best if CI was using the new `pytest` route.

run_pytest.py
kif: Because I am benchmarking it ... 
* Qt tests which randomly crash. Especially 2 of them are really brittle.
* numexpr / accelerate / mutithreaded section which crash the interpreter ... randomly
* Performance for running the tests which are far from being determinisitic.
* To be able to run the test-suite of an installed version of pyFAI without depending on a third party library, for example to debug odd behavior seldom seen.

Until those issues have been fixed, the two scripts will live side by side for some time.
The later reason makes me think the unittest test-suite will remain for a bit of time.

See also: https://github.com/silx-kit/pyFAI/issues/2948

src/pyFAI/test/profiler.py
kif: Good comment, I asked Claude to provide a similar API and the JSON file produces by this plugin (runtime, setup+teardown, memory allocation per test, ...) is very convenient for profiling work.
This does not mean there is not another pytest plugin  that does this job. Pytest-memray is a god candidate but I doubt it is usable when there are 700 tests.

I am happy that Claude does the housekeeping for me :)

kif:
Just an example of the noise introduced by pytest:

Linux x86_64 32 cores / 64 threads:
64 process with  1threads/process,  44.28s
32 process with  2threads/process,  37.08s
32 process with  1threads/process,  crash interpreter, likely in a multithreaded section #2948
16 process with  4threads/process,  39.71s
 8 process with  8threads/process,  48.96s no failure
 4 process with 16threads/process,  74.69s (2 GUI test failed)
 2 process with 32threads/process, 159.35s (4 GUI test failed)
 1 process with 64threads/process, 346.54s (4 GUI test failed)

Of course, all those tests pass with unittest.

On a mac m1 (8 core, no hyperthreading, 4 Little 4 big cores):
8 process with  1threads/process,  crash, likely numexpr
8 process with  2-8threads/process,  works
4 process with  2 thread/process, 2 GUI-tests GUI failed (same as X86_64)

Issue #2948, opened 2026-09-18
Limit the number of threads in `MultiGeometry` ?


kif:
Crashes have been noticed when running pytest in parallel.

kif:
Patch multiprocessing.cpu_count for honoring OMP_NUM_THREADS and others...

kif:
`run_pytest.py -n32 -t1` tend to crash in multigeometry tests.

kif:
Some evidences:
```
Thread 0x00007f43951f96c0 (most recent call first):
  File "/usr/lib/python3.13/multiprocessing/pool.py", line 579 in _handle_results
  File "/usr/lib/python3.13/threading.py", line 994 in run
  File "/usr/lib/python3.13/threading.py", line 1043 in _bootstrap_inner
  File "/usr/lib/python3.13/threading.py", line 1014 in _bootstrap

Thread 0x00007f43959fa6c0 (most recent call first):
  File "/usr/lib/python3.13/multiprocessing/pool.py", line 531 in _handle_tasks
  File "/usr/lib/python3.13/threading.py", line 994 in run
  File "/usr/lib/python3.13/threading.py", line 1043 in _bootstrap_inner
  File "/usr/lib/python3.13/threading.py", line 1014 in _bootstrap

Thread 0x00007f43961fb6c0 (most recent call first):
  File "/usr/lib/python3.13/selectors.py", line 398 in select
  File "/usr/lib/python3.13/multiprocessing/connection.py", line 1148 in wait
  File "/usr/lib/python3.13/multiprocessing/pool.py", line 502 in _wait_for_updates
  File "/usr/lib/python3.13/multiprocessing/pool.py", line 522 in _handle_workers
  File "/usr/lib/python3.13/threading.py", line 994 in run
  File "/usr/lib/python3.13/threading.py", line 1043 in _bootstrap_inner
  File "/usr/lib/python3.13/threading.py", line 1014 in _bootstrap

Current thread 0x00007f43969fc6c0 (most recent call first):
  File "/home/jerome/workspace-ssd/pyFAI/build_py313/lib/python3/dist-packages/pyFAI/units.py", line 155 in ne_equation
  File "/home/jerome/workspace-ssd/pyFAI/build_py313/lib/python3/dist-packages/pyFAI/geometry/core.py", line 1241 in center_array
[...]

PR #2949, opened 2026-09-19
At least 2 threads for numexpr ...

kif:
related to #2948: prevent crash when NUM_THREADS=1


kif:
@t2010 you have been waiting for it. 
Related to https://github.com/pydata/numexpr/issues/569 and https://github.com/pydata/numexpr/pull/570

ci/appveyor.yml
```diff
@@ -39,16 +39,16 @@ environment:
...
-        - PYTHON_DIR: "C:\\Python\\3.12"
-          PYTHON_ARCH: "64"
+        #- PYTHON_DIR: "C:\\Python\\3.12"
+        #  PYTHON_ARCH: "64"
 
         # Python 3.13
-        - PYTHON_DIR: "C:\\Python\\3.13"
-          PYTHON_ARCH: "64"
+        #- PYTHON_DIR: "C:\\Python\\3.13"
+        #  PYTHON_ARCH: "64"
```
t20100, line 51: Why commenting those? IMO best to remove

ci/appveyor.yml
```diff
@@ -98,7 +98,7 @@ build_script:
     - 'python -c "import silx.resources; r=silx.resources.ExternalResources(\"pyFAI\",\"http://www.silx.org/pub/pyFAI/testdata\");print(r.data_home)"'
 
     # Build
-    - "python run_tests.py -m"
+    # - "python run_pytest.py"
```
t20100, line 101:
```suggestion
```

t20100 (approved):
So tests are fixed, but parallel execution of pyfai can still seg fault and would require the fix in numexpr, right?

ci/appveyor.yml
kif: Because it will be re-shuffled in a week with the arrival of 3.15: I'll keep 3.11 and 3.15 only.

kif:
The bug in numexpr still exist, a AI agent proposed a solution but is is apparently not compatible with 3.13t/3.14t.
There is a workaround in place in this PR. Should only affect ARM architecture since it has no hyperthreading.

